PR #13, opened 2021-02-03
update Everything plugin

jjw24:


taooceros (approved):

taooceros:
What are these tests doing?

jjw24:
> What are these tests doing?

CI checks:
https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/pull/9

taooceros:
Why do we need to test it with both python 3.7 and 3.8?

Zeroto521:
It’s possible I would like to test with all kinds of python versions to ensure no problem. Py3.7 and Py3.8 have a very tiny changes.

taooceros:
But the question is that we are not testing whether the plugin is runnable with python, but whether the json format is valid, so Py3.7 and Py3.8 should not have difference.

PR #16, opened 2021-02-07
Update Dictionary to v2.2.1

harrynull:
Could you check if I packed the binaries correctly? Thanks.


harrynull:
Going to delay this a little bit. Found a bug that could issues when querying and going to investigate further.

jjw24:
No problem. Generally as long as all dlls are packed, then it's good. In all plugin projects, they should have the attribute output dll(don't remmeber the exact attribute name) to true in the csproj file.

taooceros (approved):

PR #28, opened 2021-02-17
Automatically check update

taooceros:
Automatically check update


jjw24:
can you also point debug flow to this json and see if etag need to be added as a property

taooceros:
image
It can load manifest successfully from my fork, which contains the e-tag property.

taooceros:
you can take a try for that from here "https://raw.githubusercontent.com/taooceros/Flow.Launcher.PluginsManifest/main/plugins.json"

jjw24:
first time submission, will they need to include E-Tag at all in the json?

taooceros:
> first time submission, will they need to include E-Tag at all in the json?

Well, even though they are not included, now, they will be generated once the action runs.

jjw24:
> > first time submission, will they need to include E-Tag at all in the json?
> 
> Well, even though they are not included, now, they will be generated once the action runs.

yep understood, sorry i missed out on this, so we dont need to really talk about E-Tag then. I will update the readme

updateChecker.py
```diff
@@ -0,0 +1,36 @@
...
+            break
+
+        latest_release = release_res.json()
+
+        version: str = latest_release['tag_name'].replace("v", "")
+        if(version > plugin["Version"] and len(latest_release["assets"]) > 0):
+            asset_url = latest_release["assets"][0]["browser_download_url"]
+            plugin["Version"] = version
+            plugin["UrlDownload"] = asset_url
+            plugin["E-tag"] = release_res.headers["ETag"]
```
jjw24, line 33: what happens if a new plugin without the E-tag property, is this going to be added with `plugin["E-tag"]`

taooceros: It is generated by Github, so it will definitely contain one.

jjw24 (approved):

PR #29, opened 2021-02-17
Auto update check

### LINKED ISSUE #4: Using Github Release to automatically check newest release
I think we can follow the way `Updater.cs` checking the newest update for Flow to automatically get the newest update for plugins instead of storing the direct download link in plugin.json.
This can avoid updating pluginsmanifest every time plugins update. Or checking the update is also part of the review process?

taooceros:
close #4 

This branch creates a new workflow, which will run every hours, to check the update for existing plugins list by restful api.


jjw24:
why is the entire plugins.json file changed, can we revert that.

jjw24:
remember to update readme  to indicate new plugins will need to be submitted via PR and updates will be automatic after release is published.

Zeroto521:
it is better we make a more general github action to let the developer auto push to rep(Flow.Launcher.PluginsManifest) when there any release.

Zeroto521:
if do formating gh action, one thing needs to deal that is the sequence two gh action.

taooceros:
> why is the entire plugins.json file changed, can we revert that.

That's change because I have encountered rest resource limit when I test it (60 per hours) so I add one attribute "E-tag" that used to check whether new update is released, and the check won't cause the limit. That is auto generated, and I think it won't affect the template. However, not quite sure whether the model for pluginsmanager may need to change.

taooceros:
> it is better we make a more general github action to let the developer auto push to rep(Flow.Launcher.PluginsManifest) when there any release.

That seems great, but I am not sure about what the detail should that be (change the version and release url). Maybe that can be done similarly like this, but may still need a pull request and manually check.

taooceros:
> if do formating gh action, one thing the sequence two gh action needs to deal.

I think for this one I set it will be done every hours scheduled, so it won't cause sequence issue.

Zeroto521:
> > if do formating gh action, one thing the sequence two gh action needs to deal.
> 
> I think for this one I set it will be done every hours scheduled, so it won't cause sequence issue.

using one .yml file to combine all gh action.And it is better not do it every day.

taooceros:
> using one .yml file to combine all gh action.And it is better not do it every day.

Well, why it is not better to do it every day? If we don't do it every day, how to get the latest release? For me, I would like to separate different types of actions separately, but that's my personal opinion.

Zeroto521:
formatting action should to a min size file, trim any space.

The file size will get bigger and bigger, so the `pm` would download the plugin slower and slower.

taooceros:
> formatting action should to a min size file, trim any space.
> 
> The file size will get bigger and bigger, so the `pm` would download the plugin slower and slower.

Well actually, it will only affect the manifest downloading speed, and I believe even thousands of plugins is added the file will not greater than 1MB, which is still fast for downloading. In addition, the downloading behavior happens on startup, and no wait is for that, so unless user open flow and suddenly want to install or update a plugin, they won't feel the difference.

We may still need the formatting of the json file, because we still need to add new plugins, and allow plugins stored outside github to be updated. So I think it may be hard (or possible no need) for us to trim all the space and intention.

taooceros:
> > > if do formating gh action, one thing the sequence two gh action needs to deal.
> > 
> > 
> > I think for this one I set it will be done every hours scheduled, so it won't cause sequence issue.
> 
> using one .yml file to combine all gh action.
> 
> There should be following.
> 
> 1. checking push
> 2. formating file
> 3. merge to master branch
> 
> And it is better not to do it every day.

Well actually from the testing of my fork, github finish those request quite quickly if they are not required to be updated.

Checking push and formatting, and merge will still be activated when pr is triggered.

taooceros:
I have taken a check, which the additional E-tag won't break backward compatibility.

jjw24:
> > why is the entire plugins.json file changed, can we revert that.
> 
> That's change because I have encountered rest resource limit when I test it (60 per hours) so I add one attribute "E-tag" that used to check whether new update is released, and the check won't cause the limit. That is auto generated, and I think it won't affect the template. However, not quite sure whether the model for pluginsmanager may need to change.
> 
> Github only allow 60 requests per hours without authentication, and I am not quite sure how to add that (and the code is public so token may be public or maybe we can add that through github action).

I dont think we are going to have 60 requests in an hour though, right?

taooceros:
> > > why is the entire plugins.json file changed, can we revert that.
> > 
> > 
> > That's change because I have encountered rest resource limit when I test it (60 per hours) so I add one attribute "E-tag" that used to check whether new update is released, and the check won't cause the limit. That is auto generated, and I think it won't affect the template. However, not quite sure whether the model for pluginsmanager may need to change.
> > Github only allow 60 requests per hours without authentication, and I am not quite sure how to add that (and the code is public so token may be public or maybe we can add that through github action).
> 
> I dont think we are going to have 60 requests in an hour though, right?

If we don't have the E-tag, after our plugins become more than 60, we will.

jjw24:
> it is better we make a more general github action to let the developer auto push to rep(Flow.Launcher.PluginsManifest) when there any release.

At the moment it is better to if Flow does the lifting, less set up for plugin developers, however this is not to say if we find this is not working as expected, we do not change to push from plugin projects.

taooceros:
> At the moment it is better to if Flow does the lifting, less set up for plugin developers, however this is not to say if we find this is not working as expected, we do not change to push from plugin projects.

Well the E-tag actually didn't need to be generated by developers. It will be automatically got from the restful api from github.

jjw24 (changes requested):
please update readme on:
- etag
- auto update version time interval

jjw24:
dont know what is going on here with github but looks like we are missing commits, i can see from my local git there are a lot more commits than current two

taooceros:
Seems that this from branch in my fork. It seems that I close the wrong pull request😅.

taooceros:
migrate to here #28

PR #34, opened 2021-02-19
fix update fail

jjw24:


taooceros:
What's the difference between these two? I believe the fail is because of no file has changed.

jjw24:
the actions? no diff, i am testing because i cant trigger the actions manually...

taooceros:
> the actions? no diff, i am testing because i cant trigger the actions manually...

If we just want to trigger it manually, just add one more "on" statement to allow manually trigger.

PR #50, opened 2021-06-11
CLN, CI: all scripts and configs in one

Zeroto521:
This project is used to store plugins information.
So helper script should not place at the top level of the project.
With more helper scripts come up, we better put them in another place.


ci/src/updater.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+
+    if res.status_code in (403, 304):
+        return info
+
+    latest_rel = res.json()
+    assets = latest_rel["assets"]
+    if assets:
+        info[url_download] = assets[0]["browser_download_url"]
+        info[tag] = res.headers["ETag"]
```
taooceros, line 43: Since you are making this with constant, why not make the two etag the same?

taooceros: Just change both to the one in the upstream. Etag property is only for updater so it won't affect flow.

.github/workflows/updater.yaml
```diff
@@ -6,26 +6,30 @@ on:
...
-          fetch-depth: 0 # otherwise, you will failed to push refs to dest repo
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+          # otherwise, you will failed to push refs to dest repo
+          fetch-depth: 0
+
+      - uses: actions/setup-python@v2
         with:
-          python-version: 3.8
+          python-version: "3.x"
```
taooceros, line 20: What does it mean by 3.x? Will it take the latest python 3 version?

.github/workflows/validator.yaml
```diff
@@ -1,9 +1,10 @@
...
-# Triggers the workflow on push or pull request events
-on: [push, pull_request]
-
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
 
 jobs:
```
taooceros, line 9: Since you are changing it. Do you think we need to do the validator from both py 3.7 and 3.8? I believe they will do the same job.

ci/src/updater.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+
+    if res.status_code in (403, 304):
+        return info
+
+    latest_rel = res.json()
+    assets = latest_rel["assets"]
+    if assets:
+        info[url_download] = assets[0]["browser_download_url"]
+        info[tag] = res.headers[etag]
```
taooceros, line 43: why `info[tag] = res.headers[etag]` not `info[etag] = res.headers[etag]`

Isn't tag is the raw E-tag string?

ci/src/validator.py
```diff
@@ -0,0 +1,20 @@
+# -*-coding: utf-8 -*-
+from _utils import clean, id_name, language_list, language_name, plugin_reader
+
+plugin_infos = plugin_reader()
+
+
+def test_uuid_unique():
+    uuids = [info[id_name] for info in plugin_infos]
+    uuids = [clean(uuid) for uuid in uuids]
```
taooceros, line 9: why not merge this two into one?

ci/src/validator.py
```diff
@@ -0,0 +1,20 @@
...
+
+    msg = f"The '{id_name}' is not unique."
+    assert len(uuids) == len(set(uuids)), msg
+
+
+def test_language_in_list():
+    languages = [info[language_name] for info in plugin_infos]
+    languages = [clean(language) for language in languages]
```
taooceros, line 17: same as last one

.github/workflows/updater.yaml
```diff
@@ -6,26 +6,30 @@ on:
     - cron: "0 * * * *"
```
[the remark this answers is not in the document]
taooceros: Sorry I forget to reply till now. I make it update hourly because the RestAPI limit from github is refreshed hourly.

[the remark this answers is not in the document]
taooceros: Daily seems working fine.

ci/src/validator.py
taooceros:
```suggestion
    uuids = [clean(info[id_name]) for info in plugin_infos]
```

ci/src/validator.py
taooceros:
```suggestion
def test_language_in_list():
    languages = [clean(info[language_name]) for info in plugin_infos]
```

ci/src/validator.py
Zeroto521: After that, the code would be hard to read.

You could get the meaning instantly.

And there is no performance improvement.

If there is not `info[id_name]` is `id`, there could be much better.

taooceros: Theoretically, there will be performances improvement because the previous one is allocating two array, though that can be neglected in practice.
Personally, I won't consider clean(info[id_name]) as complex inlining. However, it's not a big deal, so if you don't like that, just keep it as now.

.github/workflows/updater.yaml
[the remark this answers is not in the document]
Zeroto521: Another pr would handle this

ci/src/validator.py
Zeroto521: They are both $O(n)$.
There is only one array, no two arrays.
After the list no use anymore, this variable point to data memory would be cleaned.

.github/workflows/validator.yaml
Zeroto521: You thought from function view, I thought from test view.

This project doesn't need two validations.

taooceros: I mean we don't need two testing environment. I agree that we should only have one validation.

.github/workflows/updater.yaml
taooceros: @Zeroto521 How's going🧐?

Zeroto521:
It's ready to merge.

taooceros:
@Zeroto521 After the merge of #57 , these pr won't be tested, but the requirement still exist. Do you know how to remove that?

Zeroto521:
> @Zeroto521 @jjw24 After the merge of #57 , these pr won't be tested, but the requirement still exist. Do you know how to remove that?

discuss this at #58

jjw24:
Sorry I have not been looking at this PR, I think @taooceros you have right? Let me know if it's good to go and I will manually merge it

jjw24:
i have changed target branch and doubled checked changes are the same.

Zeroto521:
it changed to plugin_api_v2 branch too.

taooceros:
Sorry haven't got time to double check this PR. I think it is ready to merge?

taooceros (approved):

PR #56, opened 2021-07-15
CI: hourly update -> every 3h update

Zeroto521:
https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/actions/workflows/updater.yaml

Hourly update upstream is too frequent.
Let us change to a daily update.

### DIFF
--- a/.github/workflows/updater.yaml
+++ b/.github/workflows/updater.yaml
@@ -3,7 +3,7 @@ name: Auto Update
 on:
   workflow_dispatch:
   schedule:
-    - cron: "0 * * * *"
+    - cron: "0 0 * * *"
 
 jobs:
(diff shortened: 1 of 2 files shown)


taooceros (approved):

jjw24 (changes requested):
Maybe every 3 hrs instead? Daily means plugin fix is updated in their repo but the users won't be able to download until next day, I think it's too long. 

Also need to update readme as well to reflect this change

jjw24:
Just need to update this line in the readme:
```
Our CI will automatically check for new updates from plugins at every hour and update to the newer version if they are stored in the Github Release. 
```

jjw24 (approved):

PR #57, opened 2021-07-15
CI: check running when plugins.json get changed

Zeroto521:
[validator](https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/blob/main/tests.py) is running only for [plugins.json](https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/blob/main/plugins.json).

But we do other things, this ci would also run. That's unnecessary.

### DIFF
--- a/.github/workflows/validator.yaml
+++ b/.github/workflows/validator.yaml
@@ -1,9 +1,12 @@
 name: Plugins manifest validation
 
-
-# Triggers the workflow on push or pull request events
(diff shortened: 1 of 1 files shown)


jjw24:
is there a way to set it so the pipeline status check will also be bypassed if it's not touching the plugins file?

jjw24 (approved):

PR #59, opened 2021-07-16
feat: add a new plugin could show system time

Zeroto521:
Plugin demo
demo

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -238,5 +238,16 @@
         "UrlDownload": "https://github.com/Ma-ve/Flow.Launcher.Plugin.FancyEmoji/releases/download/1.0.1/Flow.Launcher.Plugin.FancyEmoji.zip",
         "UrlSourceCode": "https://github.com/Ma-ve/Flow.Launcher.Plugin.FancyEmoji/tree/master",
         "E-tag": "W/\"dfe66c7a7a2dd4c57d8b361135a6c042b25cc1e49d7ad4b9f966300d227ca65c\""
+    },
+    {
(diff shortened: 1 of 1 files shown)


plugins.json
```diff
@@ -238,5 +238,16 @@
...
+    },
+    {
+        "ID": "84ddc2a6d43911e8a8d5f2801f1b9fd1",
+        "Name": "Timestamp",
+        "Description": "Show system time and could copy that",
+        "Author": "Zero <Zeroto521>",
+        "Version": "1.0.7",
+        "Language": "python",
+        "Website": "https://github.com/Zeroto521/Flow.Launcher.Plugin.Timestamp",
+        "UrlDownload": "https://github.com/Zeroto521/Flow.Launcher.Plugin.Timestamp/archive/refs/heads/master.zip",
```
Zeroto521, line 250: Note: I don't use github release. That could be a bug to the updater.

taooceros: I remember I check whether the UrlDownload is from Github Release so it shouldn't be a problem.

Zeroto521: I have already considered using https://github.com/PyGithub/PyGithub to request GitHub after #50 merged.

jjw24: > You are right, we should add a fix to that before merging this.

should be fine to merge, @Zeroto521 can change to github release or we use PyGithub.

out of curiosity, @Zeroto521 why arent you using Github Release?

jjw24 (approved):

plugins.json
Zeroto521: No more special reason, just haven't at present.

PR #62, opened 2021-07-19
Use Skip action instead of path skip

taooceros:
https://github.com/fkirc/skip-duplicate-actions

### DIFF
--- a/.github/workflows/validator.yaml
+++ b/.github/workflows/validator.yaml
@@ -2,14 +2,23 @@ name: Plugins manifest validation
 
 on:
   push:
-    paths:
-      - "plugins.json"
   pull_request:
-    paths:
-      - "plugins.json"
 
 jobs:
+  skip_check_job:
+    runs-on: ubuntu-latest
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@master
+        with:
+          # All of these options are optional, so you can remove them if you are happy with the defaults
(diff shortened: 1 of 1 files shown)


.github/workflows/validator.yaml
```diff
@@ -2,14 +2,23 @@ name: Plugins manifest validation
 
 on:
```
Zeroto521, line 3: let simplify a bit

```yml
on: [push, pull_request]
```

.github/workflows/validator.yaml
```diff
@@ -2,14 +2,23 @@ name: Plugins manifest validation
...
 jobs:
+  skip_check_job:
+    runs-on: ubuntu-latest
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@master
+        with:
+          # All of these options are optional, so you can remove them if you are happy with the defaults
```
Zeroto521, line 16: this note looks like nothing about following, just remove it.

.github/workflows/validator.yaml
```diff
@@ -2,14 +2,23 @@ name: Plugins manifest validation
...
+          skip_after_successful_duplicate: 'true'
+          paths: '["**/plugin.json"]'
```
Zeroto521, line 18: There should more specific.
`**/plugin.json` could be under any folder.
`plugin.json` is fine

Zeroto521 (changes requested):

Zeroto521 (approved):

PR #65, opened 2021-07-23
add Cppreference plugin

peterschussheim:
Greetings!

This pull request adds my cppreference plugin.

Note that my IDE automatically formatted the entire json file.  If that is unacceptable, let me know and I will roll it back.

Thanks!


plugins.json
```diff
@@ -1,284 +1,295 @@
...
-        "Version": "1.0.0",
-        "Language": "python",
-        "Website": "https://github.com/deefrawley/Flow.Launcher.Plugin.GenConvert",
-        "UrlDownload": "https://github.com/deefrawley/Flow.Launcher.Plugin.GenConvert/releases/download/v1.0.0/Flow.Launcher.Plugin.GenConvert.zip",
-        "UrlSourceCode": "https://github.com/deefrawley/Flow.Launcher.Plugin.GenConvert/tree/main",
-        "": "W/\"64cb60a1701f6cc788997021e82b53591bf15da49533918f67df5452ff41b865\""
-    }
+  {
+    "ID": "4579F00804F94B54B23CF207856E495F",
+    "Name": "Google Search Plus",
```
Zeroto521, line 4: It seems you use 2 spaces to indent.
We use 4 spaces to indent.

Zeroto521 (changes requested):
Please fix the conflict first.

plugins.json
taooceros: I don't think format is a problem since the updater will change the format everytime it runs.

PR #77, opened 2021-10-10
fix E-tag issue

taooceros:
The current key has been "", so we can see a bunch of empty key in the plugins.json.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -81,7 +81,7 @@
         "Website": "https://github.com/Flow-Launcher/Flow.Launcher.Plugin.Everything",
         "UrlDownload": "https://github.com/Flow-Launcher/Flow.Launcher.Plugin.Everything/releases/download/v1.5.6/Flow.Launcher.Plugin.Everything.zip",
         "UrlSourceCode": "https://github.com/Flow-Launcher/Flow.Launcher.Plugin.Everything/tree/master",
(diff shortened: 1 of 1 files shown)


jjw24:
@taooceros can u quickly resolve the conflicts then I will approve and merge

Zeroto521:
I already notice this problem and do this in my branch, but to fix this needs to merge quickly.

I think this problem approves that the `"ETag"` label could be **removed**.
The script requires upstream's information, the plugins.json file content is changed then.
git would use `diff` to compare the changing content.

If there has `"ETag"`, what will happen?
The network could be a little weird sometimes.
Once the upstream plugin version doesn't changed, the `ETag` in the download URL requesting could change.
So Etag was changed, we have to update plugins.json. But this commit isn't necessary.

taooceros:
https://docs.github.com/en/rest/guides/getting-started-with-the-rest-api#conditional-requests

Though I think it is possible to add a api token to do oauth (which will increase the limit to 1000), but why not use etag to avoid duplicate check.

taooceros:
> https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/blob/e1c640d29cfbb503c6125861d75602c210404f45/.github/workflows/updater.yaml#L5-L6
> 
> basic we require **once every 3 hours**, not frequently.

The limit is 60 per hours without oauth

Zeroto521:
Well, in that way, let quickly fix this.

PR #83, opened 2021-10-31
use cdn instead of direct raw github icon path

taooceros:


taooceros:
This will greatly help chinese user to load icon who don't use special setting for github.

onesounds (approved):

Garulf:
How is this URL generated?

jjw24:
They are CDN provider, check their website

taooceros:
> How is this URL generated?

Replace github with their website by using str replace (regex command in nushell).

PR #88, opened 2021-11-08
update readme IcoPath explanation

jjw24:


jjw24:
hmm @Zeroto521 @taooceros i thought the skip_check_job should skip modifications not made to plugins.json, doesnt seem to be the case:
https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/runs/4145198764?check_suite_focus=true

Garulf:
At this point I'm wondering if we should just have devs add their repo URL to a CSV file and have an action on PR fetch the details.

jjw24:
> At this point I'm wondering if we should just have devs add their repo URL to a CSV file and have an action on PR fetch the details.

This is actually with considering, just pull the data from the repo url provided.

Garulf:
I'm doing some research & testing on this to see if its possible.

jjw24:
> I'm doing some research & testing on this to see if its possible.

Although it's a 2 min job to copy and paste that information from plugin.json

Garulf:
> > I'm doing some research & testing on this to see if its possible.
> 
> Although it's a 2 min job to copy and paste that information from plugin.json

True. But it leaves a lot of room for errors and its one of those situations where all the info is already there and available to grab so why not stream line it?

taooceros:
Add use cdn is recommended may be good

Garulf:
> Add use cdn is recommended may be good

See https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/pull/88/commits/76355a807cad2d054b9317d3230dd75d459bdc2d

Garulf (approved):

Garulf:
Seems to be hung up by older version of python-plugin workflow...

PR #89, opened 2021-11-09
Generate plugin manifest from source URL

Garulf:
This should allow an author to simple paste their source URL into a newline delimited text file and have the plugin manifest populate with required data once the PR is pushed.

Still very much work in progress


jjw24:
Is the txt file for user to paste in their url?

jjw24:
Need to add a simple validation to fail if the pasted url is not GitHub, which then they will need to manually add.

Garulf:
> Is the txt file for user to paste in their url?

Yes, if the dev pastes in their github repo URL the CI will grab all relevant information based on that URL once the PR is merged.

Garulf:
> Need to add a simple validation to fail if the pasted url is not GitHub, which then they will need to manually add.

I can add a separate workflow that fails if the URL is not valid on PR submission. This would only run if the PR changes `plugins.txt`

jjw24:
we still support non-github submissions though, so maybe failure is not ideal. Maybe skip?

PR #90, opened 2021-11-11
Test python plugins on PR

Garulf:
This will test if a Python plugin that was submitted runs without errors. This will make sure a plugin will run as packaged without any extra installation.


deefrawley:
I'll let one of the CI gurus approve but this will be great to have and stop broken python plugins getting into prod (especially while we get over the hump of the 'new' way to include dependencies)

Garulf:
> I'll let one of the CI gurus approve but this will be great to have and stop broken python plugins getting into prod (especially while we get over the hump of the 'new' way to include dependencies)

Exactly my intention! Thanks for taking a look.

.github/workflows/test-python.yml
```diff
@@ -0,0 +1,39 @@
+name: Test run Python plugins
+on:
+  workflow_dispatch:
+  pull_request:
+    types: [opened, edited, reopened, ready_for_review]
+    paths:
+      - 'plugins.json'
+jobs:
+  test:
```
jjw24, line 9: We ready have a ci job called test I think, so can we rename this to something else so I can also make it a requirement for merging

Garulf: Changed to `test-plugin`

Garulf:
Now checks for the bottom most plugin without a `Tested` Boolean field.

Garulf:
Okay, the python script now runs the plugin and checks it's return code. If the return code is non-zero it fails, otherwise we commit `true` to `Tested` field in the plugin manifest on in the PR.

Garulf:
Should be ready to rock, tested numerous times.

.github/workflows/test-python.yml
```diff
@@ -0,0 +1,33 @@
+name: Test run Python plugins
+on:
+  workflow_dispatch:
+  pull_request:
+    types: [opened, edited, reopened, ready_for_review]
+    paths:
+      - 'plugins.json'
+jobs:
+  test-plugin:
```
jjw24, line 9: sorry, can we change to 'test-run-plugin'? more descriptive

Garulf: Sure, done.

jjw24:
> Tested field isnt updated in the manifest until the PR is accepted and merged.

To clarify, job will fail if non-zero exit code right? and will update the tested field when plugin submission is merged in?

Garulf:
> > Tested field isnt updated in the manifest until the PR is accepted and merged.
> 
> To clarify, job will fail if non-zero exit code right? and will update the tested field when plugin submission is merged in?

Yes, if the plugin fails to run it'll return a non-zero exit and fail the test. If we accept the plugin PR it should be picked up on merge by `python-passed.yml` which adds a `Tested` field with a True` value to any python plugin missing one (which should be any plugin that got accepted).

jjw24 (approved):

jjw24:
@Garulf got some failures in the ci run. Are you able to look into?

Garulf:
> @Garulf got some failures in the ci run. Are you able to look into?

I see the issue.

Garulf:
@jjw24 Fixed: https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/pull/94

PR #97, opened 2021-12-11
Use commit action

Garulf:
Use the same git commit function as other workflows.


jjw24:
Do you not need to specify the token- GitHub secret to push?

jjw24:
I will merge it in but I suspect you need to use the GitHub secret with the checkout action, see updater

jjw24 (approved):

PR #104, opened 2021-12-12
update python plugin tests 

jjw24:
tested python plugins, update the plugin list.


jjw24:
@Garulf i removed the 'Tested' field so it can re-test, but seems to indicate no issues with plugins, even with RollDice. Have I not done the test correctly?

Garulf:
After staring at the same code for 30 minutes I finally noticed what was missing. Fixed in https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/pull/105

Garulf:
I should mention I never imagined multiple plugins being submitted at once. The Python testing only runs on the first plugin without a tested field.

jjw24:
Nah, keep it how it is should be fine, I'm just trying to find an easier way to test run the current plugins instead of manually download and run each one.

Garulf:
Hmm.. encoding error failure?

Garulf:
Hmm I wonder if we could winget flow launcher in the CI. Should fix Flox erroring out.

Garulf:
Sorry hijacked to get Flox working

Garulf (approved):

PR #105, opened 2021-12-12
Use python exe

Garulf:
Need to specify file is run with Python

### DIFF
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -14,7 +14,7 @@ jobs:
           paths: '["plugins.json"]'
   test-run-plugin:
     needs: skip_check_job
-    if: ${{ needs.should_skip.outputs.should_skip != 'true' }}
+    if: ${{ needs.skip_check_job.outputs.should_skip != 'true' }}
     runs-on: windows-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


Garulf:
Odd it's not skipping Python test when `plugins.json` is *not* changed

Garulf:
Ah, validator has the same typo which i copied and pasted...

Garulf:
Successfully skip if plugins.json isn't changed 🤞

.github/workflows/test-python.yml
```diff
@@ -5,16 +5,18 @@ jobs:
...
-        with:
-          skip_after_successful_duplicate: 'true'
-          paths: '["plugins.json"]'
+    - uses: actions/checkout@v2
+    - uses: dorny/paths-filter@v2
+      id: filter
+      with:
+        filters: |
+          plugins:
+            - 'plugins.json'
```
taooceros, line 16: Why change to this?

Garulf: Not sure exactly why but the old workflow kept detecting changes to plugins.json regardless if it was changed in PR or not.

jjw24:
awesome, thanks for this. will test this in the other pr before merging in.

Garulf:
It's not breaking anything but validator is probably having the same issue.

jjw24 (approved):

PR #109, opened 2021-12-16
Python Plugin Testing upgrade

Garulf:
* Cleans up code
* Doc strings and type hints
* Verify JSON output
* Check and fail on multiple Plugin submission
* Setup faux Flow Launcher structure
* Prioritize `UrlDownload` field for release testing (Fallback to github release if available)
* Support Flox
* Parse `SettingsTemplate.yaml` and send default values to Plugin

### DIFF
--- a/ci/src/test-python.py
+++ b/ci/src/test-python.py
@@ -63,7 +63,7 @@ def get_latest_plugin(manifest: dict) -> dict:
                 sys.exit(0)
             break
     else:
-        print("No Untested plugin found!")
+        print_section("Test failed!", "No Untested plugin found!\nTest could not find a plugin without \"Tested\" key.")
         sys.exit(1)
     return _plugin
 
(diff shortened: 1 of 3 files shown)


ci/src/test-python.py
```diff
@@ -9,69 +10,187 @@
...
+def setup_flow_environment() -> None:
+    """Setup Flow Launcher local & roaming directory."""
+    _mkdir(USER_PATH)
+    _mkdir(APP_PATH)
+    for _dir in USER_DIRS:
+        _mkdir(Path(USER_PATH, _dir))
+    for _dir in APP_DIRS:
+        _mkdir(Path(APP_PATH, _dir))
+    os.makedirs(Path(USER_PATH, "Settings", "Plugins"), exist_ok=True)
+    os.makedirs(Path(APP_PATH, "app-1.9.0"), exist_ok=True)
```
jjw24, line 115: this will still hold when the version is no longer 1.9.0?

Garulf: Yeah I don't think this will affect anything.

jjw24: cool, let's use a generic version.
```suggestion
    os.makedirs(Path(APP_PATH, "app-1.0.0"), exist_ok=True)
```

jjw24 (approved):

PR #112, opened 2021-12-28
Detect and pass on non-Python detection

Garulf:
This should fix non-Python plugins from causing the test to fail.

### DIFF
--- a/ci/src/test-python.py
+++ b/ci/src/test-python.py
@@ -68,7 +68,7 @@ def get_latest_plugin(manifest: dict) -> dict:
         if _plugin["Language"] == "python" and "Tested" not in _plugin.keys():
             untested_plugins.append(_plugin)
         if _plugin["Language"] != "python" and "Tested" not in _plugin.keys():
-            print_section("Non-Python plugin detected!", f'Detected Plugin: {_plugin["name"]}\nPassing test for now...')
+            print_section("Non-Python plugin detected, test not required.", f'Detected Plugin: {_plugin["name"]}\nPassing test...')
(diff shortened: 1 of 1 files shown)


ci/src/test-python.py
```diff
@@ -67,8 +67,11 @@ def get_latest_plugin(manifest: dict) -> dict:
     for _plugin in manifest[::-1]:
         if _plugin["Language"] == "python" and "Tested" not in _plugin.keys():
             untested_plugins.append(_plugin)
+        if _plugin["Language"] != "python" and "Tested" not in _plugin.keys():
+            print_section("Non-Python plugin detected!", f'Detected Plugin: {_plugin["name"]}\nPassing test for now...')
```
jjw24, line 71:
```suggestion
            print_section("Non-Python plugin detected, test not required.", f'Detected Plugin: {_plugin["name"]}\nPassing test...')
```

ci/src/test-python.py
```diff
@@ -67,8 +67,11 @@ def get_latest_plugin(manifest: dict) -> dict:
...
+            sys.exit(0)
     if len(untested_plugins) == 0:
-        print_section("Test failed!", "No Untested plugin found!\nTest could not find a plugin without \"Tested\" key.")
+        print_section("Test failed!", "Test could not find a plugin without \"Tested\" key.")
```
jjw24, line 74:
```suggestion
        print_section("Test failed!", "The new plugin should not have the \"Tested\" key.")
```

jjw24 (approved):

PR #120, opened 2022-01-02
Add Epic launcher to manifest

Garulf:
Adds a plugin that launches your Epic Games Library.

image


Garulf:
Failing because the test environment doesn't have Epic game store and I didnt handle the exception... one moment...

jjw24 (approved):

PR #126, opened 2022-01-13
Send discord webhook on plugin updates

Garulf:
Updater will send a discord webhook to Discord when a plugin's version is changed.

Just need to provide a webhook in repo's secrets named: `DISCORD_WEBHOOK`.

image

### DIFF
--- a/ci/src/discord.py
+++ b/ci/src/discord.py
@@ -1,0 +1,34 @@
+import requests
+
+def update_hook(webhook_url: str, info: dict) -> None:
+    embed = {
+        "content": None,
+        "embeds": [
+            {
+            "title": "Plugin Updated",
+            "description": f"{info['Name']} plugin has been updated to v{info['Version']}!",
+            "url": info['Website'],
+            "color": None,
+            "fields": [
+                {
+                "name": "Author",
+                "value": info['Author'],
+                "inline": True
+                },
+                {
+                "name": "Language",
+                "value": info['Language'],
+                "inline": True
+                },
+                {
+                "name": "Download",
+                "value": info['UrlDownload']
+                }
+            ],
+            "thumbnail": {
+                "url": info['IcoPath']
+            }
+            }
+        ]
+        }
+    requests.post(webhook_url, json=embed)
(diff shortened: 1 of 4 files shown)


ci/src/discord.py
```diff
@@ -0,0 +1,34 @@
...
+                "value": info['UrlDownload']
+                }
+            ],
+            "thumbnail": {
+                "url": info['IcoPath']
+            }
+            }
+        ]
+        }
+    requests.post(webhook_url, json=embed)
```
taooceros, line 34: We don't need to serialize here?

ci/src/updater.py
```diff
@@ -30,19 +33,30 @@ def batch_github_plugin_info(info: P, tags: ETagsType) -> P:
     if assets:
         info[url_download] = assets[0]["browser_download_url"]
         info[version] = clean(latest_rel["tag_name"], "v")
+
+    if info[version] != clean(latest_rel["tag_name"], "v"):
```
taooceros, line 37: < or > may be a better check?

Garulf: Done

ci/src/discord.py
Garulf: Sorry, what do you mean?

taooceros: nvm I check the docs

ci/src/updater.py
```diff
@@ -30,19 +33,30 @@ def batch_github_plugin_info(info: P, tags: ETagsType) -> P:
     if assets:
         info[url_download] = assets[0]["browser_download_url"]
         info[version] = clean(latest_rel["tag_name"], "v")
+
+    if version_tuple(info[version]) < version_tuple(latest_rel["tag_name"]):
+        tqdm.write(f"Update detected: {info[id_name]} {info[version]}")
+        try:
+            update_hook(webhook_url, info)
+        except:
+            # We dont want to crash if the webhook is not working
```
jjw24, line 42: do we not want it to error rather than fail silently?

Garulf: I was thinking of adding output to console but I dont think it's that important. What we really dont want to happen is have the whole script fail because of a webhook.

jjw24: how do we know if the webhook is working or not though, plugin updates are not often, maybe write a console output before passing?

Garulf: I'll add a `bool` and check at the end of loop to print an error message.

jjw24: why not just write the output above the pass?

Garulf: Prints message out on error

taooceros (approved):

PR #152, opened 2022-04-18
Notion search

danielduckworth:
Added requests to requirements so it should work by installing from Flow Launcher now.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -820,5 +820,17 @@
         "UrlDownload": "https://github.com/LeoDupont/Flow.Launcher.Plugin.DateDiff/releases/download/v1.0.0/Flow.Launcher.Plugin.DateDiff.zip",
         "UrlSourceCode": "https://github.com/LeoDupont/Flow.Launcher.Plugin.DateDiff/tree/main",
         "IcoPath": "https://cdn.jsdelivr.net/gh/LeoDupont/Flow.Launcher.Plugin.DateDiff@main/images/icon.png"
+    },
+    {
+        "ID": "3e268e4d-50a7-4847-aa58-adb7b743de1a",
+        "Name": "Notion Search",
+        "Description": "Search notion pages and databases",
+        "Author": "Duckworth",
+        "Version": "0.0.4",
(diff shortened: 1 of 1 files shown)


jjw24:
The CI error just reminded me, I forgot to mention previously, it might be good to handle unauthorised exception thrown since the hotkey is `n` and you can run into this error if installed for the first time and not yet added the token.

danielduckworth:
> The CI error just reminded me, I forgot to mention previously, it might be good to handle unauthorised exception thrown since the hotkey is `n` and you can run into this error if installed for the first time and not yet added the token.

Yep, good idea. Will do.

jjw24:
You can return a result when a token is not set to let users know that they need to go to settings to add it.

Garulf:
I ran into a few issues when running the plugin. I'll submit a PR with a fix in a moment.

danielduckworth:
> I ran into a few issues when running the plugin. I'll submit a PR with a fix in a moment.

Hi @Garulf. I've added two try/except flows to handle an empty token setting and queries that return API errors. However the latter catches cases where the API returns no results without errors, yet the message to the user suggests there was an error. So I need to catch the actual error here to handle it more accurately.

The other problem is that in Flow Launcher, if I enter the action keyword (n) but do not enter a search term and wait, eventually it searches with an empty query which crashes because the helper function makes assumptions about the title key in the dictionary. Seems there might be a variation of the page result where the name key isn't used or used differently. I was thinking it might be better to enumerate the whole dictionary to return the values for the title keys rather than a fixed key/value pair which could change as the API endpoint changes.

jjw24:
Hey @danielduckworth, any updates and ready to get published?

danielduckworth:
Hey @jjw24 I've updated the plugin. I messed up the GitHub process and have accidently created a new pull requesting rather than resolving this. Not sure what happened.

https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/pull/155

danielduckworth:
I'll try again

PR #161, opened 2022-06-22
New Unity3D helper plugin

falldeaf:
A plugin for Flow launcher that let's you list, search, and open your Unity projects with the correct editor version. The context menu will also allow you to view the project in VSCode, Github Desktop, or open the projects Assets folder in explorer.


falldeaf:
> @jjw24 I wonder should we add a readme page when installing new plugin (or at least direct users to the reader page)? There're several plugins that requires some additional setup like this.

That's a great idea! That would definitely be useful for users

jjw24:
> @jjw24 I wonder should we add a readme page when installing new plugin (or at least direct users to the reader page)? There're several plugins that requires some additional setup like this.

Yes absolutely and also given non .Net and python plugins still need manual setup.

I am thinking to streamline the process, let flow manage opening the instructions page by adding an optional field in plugin.json 'instructionsUrl'. When not blank, flow will prompt user on startup plugin load once, and when will open the url when user clicks yes.

taooceros (approved):

PR #162, opened 2022-07-01
Update NumberConverter plugin to 1.0.2

liberize:


jjw24:
Hey how's it going, plugin version updates are automatic and will occur at an interval after you do a release, so no need to update it here.

PR #165, opened 2022-08-02
Search npm plugin

gabrielcarloto:


gabrielcarloto:
I found that for some reason the dependencies modules were not included in the plugin `.zip` file. I will fix this.

gabrielcarloto:
I forgot to mention it but I fixed that problem in update 1.0.1.

taooceros (approved):

PR #168, opened 2022-08-06
Added a CyberChef plugin

gamous:
* Added a CyberChef plugin

* Update plugins.json

[Cyberchef](https://github.com/gchq/CyberChef) is a web-based, feature-rich coding suite framework. My plugin allows users to quickly send queries to a custom built CyberChef page efficiently.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -919,6 +919,6 @@
         "Website": "https://github.com/gamous/Flow.Launcher.Plugin.CyberChef",
         "UrlDownload":"https://github.com/gamous/Flow.Launcher.Plugin.CyberChef/releases/download/v1.0.0/Flow.Launcher.Plugin.CyberChef.zip",
         "UrlSourceCode": "https://github.com/gamous/Flow.Launcher.Plugin.CyberChef",
-        "IcoPath": "https://github.com/gamous/Flow.Launcher.Plugin.CyberChef/raw/main/Images/app.png"
+        "IcoPath": "https://cdn.jsdelivr.net/gh/gamous/Flow.Launcher.Plugin.CyberChef@main/Images/app.png"
       }
 ]
--- a/plugins.json
+++ b/plugins.json
@@ -908,5 +908,17 @@
         "UrlDownload": "https://github.com/Echostorm44/FlowLauncherPluginEmailTo/releases/download/v1.0.0/Flow.Launcher.Plugin.FlowLauncherPluginEmailTo.zip",
         "UrlSourceCode": "https://github.com/Echostorm44/FlowLauncherPluginEmailTo",
         "IcoPath": "https://cdn.jsdelivr.net/gh/Echostorm44/FlowLauncherPluginEmailTo@main/Flow.Launcher.Plugin.FlowLauncherPluginEmailTo/icon.png"
-    }
+    },
+    {
+        "ID": "188b0b5c-159b-11ed-861d-0242ac120002", 
+        "Name": "CyberChef",
+        "Description": "Send your input to the CyberChef",
+        "Author": "Gamous",
+        "Version": "1.0.0",
+        "Language": "python",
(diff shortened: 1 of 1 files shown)


jjw24 (commented):
Hi @gamous how's it going? I tried your plugin, when the website opens it looks like the query string has been passed through but then the site seems to have reloaded and the string was lost. It works however if I hardcoded in the url into main.py

gamous:
> how's it going? I tried your plugin, when the website opens it looks like the query string has been passed through but then the site seems to have reloaded and the string was lost. It works however if I hardcoded in the url into main.py

I'm sorry I wasn't aware of this bug. I tested the plugin with self-hosted CyberChef set up, so I didn't notice a type error in determining the fallback to the default site.
This issue has been fixed and the release of my plugin repository updated. Maybe you can try again.

jjw24:
Hi @gamous , I am still getting the same behaviour as before after downloading the new updated zip. You can test as well with `pm install https://github.com/gamous/Flow.Launcher.Plugin.CyberChef/releases/download/v1.0.0/Flow.Launcher.Plugin.CyberChef.zip`

gamous:
Hi~ @jjw24,thank you for your patience. I fixed the behavior of the contextmenu and then used this command to uninstall and install several times to test. I'm sure this plugin works now.

Garulf:
Hello @gamous!

Is the newest version still 1.0.0?

If not the manifest still points to the older (non-working?) version.

gamous:
Hi @Garulf ～ Yes. The newest version is still 1.0.0. I delete older and reuse the same version tag to release current zip. I thought it was reasonable to remove release that didn't work, without having to change my manifest pr submission. In other words, the submission in this pr still points to a well-worked version now.

jjw24:
Hi @gamous, I just tested the plugin, running the query `cf hello`, still the same result, website opens but still opens up as https://gchq.github.io/CyberChef/. Have I missed something?

gamous:
Hi~ @jjw24. I am ashamed of this bug. You're right, it's a function issue and I don't notice. In fact, I noticed early on that the problem was URL redirection due to "=" so strip was added to remove the "=" at the end. However, due to an error, I did not find that the wrong version was submitted. I tested the release using an integer multiple of the payload so that no problem was triggered. The release file is updated.

jjw24 (approved):

plugins.json
```diff
@@ -908,5 +908,17 @@
...
+        "ID": "188b0b5c-159b-11ed-861d-0242ac120002", 
+        "Name": "CyberChef",
+        "Description": "Send your input to the CyberChef",
+        "Author": "Gamous",
+        "Version": "1.0.0",
+        "Language": "python",
+        "Website": "https://github.com/gamous/Flow.Launcher.Plugin.CyberChef",
+        "UrlDownload":"https://github.com/gamous/Flow.Launcher.Plugin.CyberChef/releases/download/v1.0.0/Flow.Launcher.Plugin.CyberChef.zip",
+        "UrlSourceCode": "https://github.com/gamous/Flow.Launcher.Plugin.CyberChef",
+        "IcoPath": "https://github.com/gamous/Flow.Launcher.Plugin.CyberChef/raw/main/Images/app.png"
```
jjw24, line 922:
```suggestion
        "IcoPath": "https://cdn.jsdelivr.net/gh/gamous/Flow.Launcher.Plugin.CyberChef@main/Images/app.png"
```

jjw24 (approved):

jjw24:
> Hi~ @jjw24. I am ashamed of this bug. You're right, it's a function issue and I don't notice. In fact, I noticed early on that the problem was URL redirection due to "=" so strip was added to remove the "=" at the end. However, due to an error, I did not find that the wrong version was submitted. I tested the release using an integer multiple of the payload so that no problem was triggered. The release file is updated.

All good, no stress :)

PR #175, opened 2022-08-16
Added my new IP plugin

Echostorm44:
added new plugin for just getting 1 public and 1 real private IP

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -945,5 +945,18 @@
         "UrlDownload": "https://github.com/qjcXu/Flow.Launcher.Translation/releases/download/v0.1.0/Flow.Launcher.Translation.0.1.0.zip",
         "UrlSourceCode": "https://github.com/qjcXu/Flow.Launcher.Translation",
         "IcoPath": "https://cdn.jsdelivr.net/gh/qjcXu/Flow.Launcher.Translation/src/assets/imgs/translate.png"
+    },
+    {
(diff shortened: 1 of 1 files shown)


taooceros:
Just a bit curious, do we have any functionality difference with this one? https://github.com/taooceros/Flow.Plugin.IPAddress

Echostorm44:
Yeah, theirs just pulls all the local IPs including virtual ones from stuff like HyperV, VMware, docker etc.  They also don't pull your public IP.  Mine only pulls your real LAN IP and your public IP.

taooceros:
> Yeah, theirs just pulls all the local IPs including virtual ones from stuff like HyperV, VMware, docker etc so you can't tell whats what. They also don't pull your public IP. Mine only pulls your real LAN IP and your public IP.

Sure thanks. Though as what I remember that plugin also trying to fetch the public IP (I think something is not working correctly though). Do you mind adding your code in that plugin instead of adding a new plugin for a similar purpose? I feel like that may create some confusion for user to choose from. (It's totally up to you. I will be happy to merge this one).

Echostorm44:
If it's all the same to you I'd rather keep mine so that I can maintain it and keep the functionality cleaner than the other one.

taooceros:
Sure thing

taooceros (approved):

PR #176, opened 2022-08-17
update to use async as requested

Echostorm44:


taooceros:
The update is automatic. No need for another pr.

PR #178, opened 2022-09-02
Translation plugin nodejs

cenyG:


taooceros:
We already have a libretranslate plugin. Would you share the difference with that?

cenyG:
> We already have a libretranslate plugin. Would you share the difference with that?

In those implementation we have to specify `<from>`, `<to>` lang like that:
`tr <from (code)> <to (code)> <text>`,
and it's a bit lazy for me to type it all the time.

So in my case I can set `<from>` lang once, and then use just
`tr <text>`

Also we use different instances of Libretranslate.

plugins.json
```diff
@@ -960,6 +960,18 @@
         "IcoPath": "https://cdn.jsdelivr.net/gh/Echostorm44/FlowLauncherPluginMyIPs@main/Flow.Launcher.Plugin.FlowLauncherPluginMyIPs/icon.png"
     },
     {
+        "ID": "5423df3c6dd84fc5ad3ca1bfadad872dfca307c0a013d116dc3e8bed60b8ca5b",
+        "ActionKeyword": "tr",
```
jjw24, line 964:
```suggestion
```

plugins.json
```diff
@@ -960,6 +960,18 @@
...
+        "Author": "cenygg",
+        "Version": "1.0.0",
+        "Language": "executable",
+        "Website": "https://github.com/cenyG/flow-launch-translate-plugin",
+        "ExecuteFileName": "node.bat",
+        "IcoPath": "https://cdn.jsdelivr.net/gh/cenyG/flow-launch-translate-plugin@main/Images/liber_icon.png"
+     },
+     {
         "ID": "F8775498-BD72-4ECA-893F-B33863D81CB5",
         "ActionKeyword": "u",
```
jjw24, line 976:
```suggestion
```

jjw24:
Hi @cenyG, the action keyword 'tr' to trigger this plugin is duplicated with another plugin, we haven't yet implemented the ability to use duplicate action keyword which means when two plugins of the same keyword exists, one of them will not get triggered. Would you mind changing your action keyword to something else please.

cenyG:
Hi @jjw24, I have changed keyword to `tt`. 
Also made some change to main repo. Now this plugin uses Google Translate for translation, also add flag `r` for reversing `to` -> `from` translation

jjw24 (approved):
👍 

Please update readme to show user how to download nodejs to run the plugin.

PR #184, opened 2022-10-17
Remove Spell Checker Plugin

Garulf:
Due to a dependency that requires a C extension this plugin cant be packaged to work on multiple platforms.


jjw24:
Cant be packaged to work on multiple platforms, multiple platforms of windows?

jjw24:
We only support 64bit though, and people can use our python installation can't they?

Garulf:
> We only support 64bit though, and people can use our python installation can't they?

I usually try to strive for Python 3.8-current compatibility. Id like to take the plugin down until so can find a better way without C extension dependency.

PR #202, opened 2023-01-25
Add KeePass plugin

vladislav-atakhanov:


Garulf:
I'm iffy with keeping the database always unlocked as this can invite an "evil maid" type attack. 

https://www.google.com/search?q=evil+maid+attack&ie=UTF-8&oe=UTF-8&hl=en-us&client=safari

vladislav-atakhanov:
> I'm iffy with keeping the database always unlocked as this can invite an "evil maid" type attack.
> 
> https://www.google.com/search?q=evil+maid+attack&ie=UTF-8&oe=UTF-8&hl=en-us&client=safari

I don't think this is a plugin issue. The user can set a password to enter the system. I also saw that some plugins store access tokens in the Flow.Launcher

Garulf:
Looks like lxml, one of pykeepass's dependencies uses C extensions and may not work properly on Windows if your CI OS is Linux.

vladislav-atakhanov:
I fixed it. Run the tests again

vladislav-atakhanov:
I can't fix it

Garulf:
I'll take a look when I can.

Garulf:
Okay I finally got a chance to look at why it may be failing still. Looks like its the same issue as before. the `--no-binary` flag simple tells pip to ignore pre-built wheels and compile them at install on the system.

In short lxml is still using binary files not compatible with windows.

jjw24 (changes requested):
Hi @vladislav-atakhanov, Flow unfortunately does not have anything to secure passwords on its own right now. 

Currently your plugin is storing the master password to the database of KeePass as plaintext, unlike tokens this could lead to a user's very sensitive information such as banking details be at risk, especially if the user moves flow from computer to computer.

Could you implement a mechanism to store the password in Window's Credential Manager instead please. You can use libraries for example like this one https://github.com/jaraco/keyring or this https://gist.github.com/exhuma/a310f927d878b3e5646dc67dfa509b42

Garulf:
Closing for now. Please see @jjw24's message above and feel free to reopen if needed.

PR #203, opened 2023-02-06
Add hacker news plugin

Joehoel:


onesounds (dismissed):

onesounds:
@Joehoel  Sorry for the late approval. I tested it and it works well.

1. I hope there's a simple icon. as you know, Currently, the default icon that appears when no icon is present.



2. Depending on the setting, it would be good to see a specific page right away. I only look at the front page, but I think it would be good to have a setting that shows the front page right away when you enter hn

Thank you for registering the plug-in.

onesounds (dismissed):

onesounds (approved):

Joehoel:
Thanks! No worries. I will add the icon and work on your suggestions!

PR #204, opened 2023-02-11
Add Vercel plugin

guilherssousa:
https://github.com/guilherssousa/flow-launcher-vercel-plugin


onesounds (approved):

onesounds:
Thank you for making the plug-in. 

+) I hope icon is not a black or white solid color.(Depending on the flow theme, icons are often not visible at all.)  but VERCEL logo is a solid color. I think it would be good to change it to an icon wrapped in a black circular like this.  

image
(icons from https://github.com/vercel/vercel)

PR #206, opened 2023-02-17
Add JetBrainsIDEProjects plugin

kenty02:
I have created a plugin to search projects in JetBrains IDE.
It should work with all IDEs as long as they are installed from Toolbox.

https://github.com/kenty02/Flow.Launcher.Plugin.JetBrainsIDEProjects


jjw24:
@kenty02 might be worth handling the exception when JetBrains IDE is not found (a result detailing what to do), otherwise user will get a flow error pop up message.

jjw24 (approved):

PR #213, opened 2023-03-02
Audio Device Selector - Version 1.0.2 - Add Settings for different display modes

attilakapostyak:


jjw24:
@attilakapostyak plugins will auto update so there is no need for a pr :)

PR #219, opened 2023-03-10
replace temp icon with proper one for youtube downloader plugin

DeepVoyager253:
replace temp icon with proper one for youtube downloader plugin


plugins.json
```diff
@@ -1344,15 +1344,15 @@
         "ID": "d58fbe34-bce7-11ed-afa1-0242ac120002",
         "Name": "Youtube Downloader",
         "Description": "A youtube downloader",
-        "Author": "LordOfTheEel",
+        "Author": "LordOfTheEel; Eyal.Br",
```
Garulf, line 1347: This intentional?

DeepVoyager253: yeah, i want to add my real name too

Garulf (approved):

plugins.json
```diff
@@ -1344,15 +1344,15 @@
...
         "Version": "v1.0.0-beta",
```
jjw24, line 1348: I wonder if this may crash flow in areas where it needs to sort or diff the different versions.

DeepVoyager111: I don't think it will, as when I update the plugin I delete other releases. I am sure this is not the best way to handle this, but it shouldn't cause any bugs. Hope this helps

DeepVoyager111: If you are referring to letters in the plugin version, it hasn't caused any bugs so far. If it has for you, please make sure to create an issue on my repo

PR #223, opened 2023-03-19
Set correct UrlDownload and IcoPath

MichielvanBeers:
Thanks for the feedback! As requested, I've set the Github workflow/Action to set the correct name of the .zip file and set the IcoPath to use jsdeliver.


jjw24 (approved):
Perfect thanks 👍

PR #229, opened 2023-04-28
[new plugin] ClipboardR v0.1.4

rainyl:
I developed a new clipboard plugin based on the previous clipboard history plugin, but it's nearly a fresh one.

Repo: https://github.com/rainyl/Flow.Launcher.Plugin.ClipboardR

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -1450,10 +1450,7 @@
     "Version": "0.1.4",
     "Language": "csharp",
     "Website": "https://github.com/rainyl/Flow.Launcher.Plugin.ClipboardR",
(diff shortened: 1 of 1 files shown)


Garulf:
Just tested and the Preview panel addition is very nice.

I did notice that your plugin's default keyword `cc` conflicts with the currency converter plugin.
Currently Flow Launcher plugins with conflicting keywords show the first plugin loaded only. This could confuse users of your plugin.

While you don't necessarily have to change it if you do not want to I'd recommend something unique if possible.

Garulf (approved):

rainyl:
Thanks for your suggestions, I just think it's simple and super convenient to use cc, I will change it :)

PR #232, opened 2023-05-05
Switched pytube implementation to yt_dlp 

DeepVoyager253:


DeepVoyager253:
small issue when doing certain actions, will open another PR when fixed.

PR #247, opened 2023-05-26
Add VolumeFlow

asmpro7:
Plugin for controlling your Sound Volume


Garulf:
Looking at the code I'm worried activating volume controls when a keyword is typed could lead to some unexpected behavior.

Garulf (approved):

PR #251, opened 2023-06-06
fix(discord): :bug: fix webhook failure when release notes too long (#250)

### LINKED ISSUE #250: Discord webhook fails to send if release notes are too long
https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/blob/19a7158a41155a82562cf22841e8b3402fc774bb/ci/src/discord.py#L42

If the release notes are too long, the embed field value will be longer than Discord's `1024` limit, and the webhook will fail to fire.

JamesNZL:
Fixes #250.

Currently, the release notes are added directly into a new embed field without checking its length.
https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/blob/19a7158a41155a82562cf22841e8b3402fc774bb/ci/src/discord.py#L42

If the release notes are longer than Discord's `1024` embed field value API limit, the webhook will fail to fire.
> This is pretty easily reached if you link issues/commits in your release notes, like semver tools often do:
  image

I've added a little formatting function that checks the length of the release notes and gracefully truncates if necessary (ie does its best to avoid markdown formatting errors by truncating entire lines, rather than half-way through a line).

Have unit tested the function.

image

:)


ci/src/discord.py
```diff
@@ -39,5 +39,27 @@ def update_hook(webhook_url: str, info: dict, latest_ver: str, release: dict) ->
...
+        embed['embeds'][0]['fields'].append({"name": "Release Notes", "value": truncate_release_notes(release.get('body', ""), 1024)})
+    requests.post(webhook_url, json=embed)
+    
+def truncate_release_notes(release_notes: str, length: int) -> str:
+    if len(release_notes) <= length:
+        return release_notes
+    
+    TRUNCATION_MESSAGE_BASE = "\n{} lines truncated..."
+    # Will definitely not have more lines than total characters
+    maximum_lines_message_length = len(release_notes)
```
JamesNZL, line 51: Whoops, this is meant to be checking the number of digits of the length, ie
```py
len(str(len(release_notes)))
```

😅

Garulf:
Thanks! I'll take a closer look as soon as I can.

ci/src/discord.py
JamesNZL: Though, I think this still works because of negative indices 🤔

JamesNZL:
Whoops, forgot to un-draft it yesterday.

Ready for review.

ci/src/discord.py
```diff
@@ -39,5 +39,28 @@ def update_hook(webhook_url: str, info: dict, latest_ver: str, release: dict) ->
         embed['embeds'][0]["author"]["icon_url"] = f"{github_url}/{github_username}.png?size=40"
     release_notes = release.get('body')
     if release_notes and release_notes.strip():
-        embed['embeds'][0]['fields'].append({"name": "Release Notes", "value": release.get('body', "")})
-    requests.post(webhook_url, json=embed)
\ No newline at end of file
+        embed['embeds'][0]['fields'].append({"name": "Release Notes", "value": truncate_release_notes(release.get('body', ""), 1024)})
+    requests.post(webhook_url, json=embed)
+    
+def truncate_release_notes(release_notes: str, length: int) -> str:
```
Garulf, line 45:
```suggestion
def truncate_release_notes(release_notes: str, length: int = MAX_BODY_LEN) -> str:
```
We should define "1024" as a constant so we don't use magic numbers. :)

Garulf (changes requested):

ci/src/discord.py
```diff
@@ -39,5 +39,28 @@ def update_hook(webhook_url: str, info: dict, latest_ver: str, release: dict) ->
...
+    if len(release_notes) <= length:
+        return release_notes
+    
+    TRUNCATION_MESSAGE_BASE = "\n{} lines truncated..."
```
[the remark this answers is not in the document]
JamesNZL: I like the 'show more' idea; it means less will be displayed in the embed itself but it's a lot more UX-friendly.

There *is* the link to the release in the embed title, but I think it is justifiable to use the characters to have a more semantic link in the event of truncation.

ci/src/discord.py
JamesNZL: Yup good catch 😅, I think I let it slip cause I figured the surrounding logic was pretty tightly-coupled to the API anyway—will fix though

Garulf: No problem. This doesn't affect the code as much but helps readability later.

JamesNZL:
Ready for re-review 🙂

Garulf (approved):

PR #254, opened 2023-06-11
Add link checker github action

JohnTheGr8:
Added a Github action that checks every link in the `plugins.json` file whenever the file is updated (push or PR).

The action also runs on a cron schedule so that we get alerted if one of the plugin repositories goes down.

I've also fixed 3 links that were broken.


Garulf:
I have similar check added in my PR https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/pull/246

JohnTheGr8:
> I have similar check added in my PR #246

sorry I didn't notice... but regardless I think using a dedicated tool for this is a more elegant solution ;)

Garulf:
> > I have similar check added in my PR #246
> 
> sorry I didn't notice... but regardless I think using a dedicated tool for this is a more elegant solution ;)

The solution is good, however I'd like to check URLs on submission as well. 

I am for whatever works best.

JohnTheGr8:
> The solution is good, however I'd like to check URLs on submission as well.

it does, I made it run on push, PR and on a schedule. Am I missing something? (not very familiar with github actions)

Garulf:
> also runs on a cron schedule

Sorry I misread and thought it only ran on cron

JohnTheGr8:
> Sorry I misread and thought it only ran on cron

I mean the check did run on this PR 😂

Garulf:
Will this fail a submission if someone else's link is dead?

JohnTheGr8:
> Will this fail a submission if someone else's link is dead?

I think we can configure what checks must be green before a PR can be merged...

Garulf:
> I think we can configure what checks must be green before a PR can be merged...

IIRC its all or nothing with github tests

JohnTheGr8:
> IIRC its all or nothing with github tests

well, shouldn't be an issue, there's just 3 dead links after all this time...

Garulf (approved):

PR #255, opened 2023-06-17
Update plugins.json

DiekoMA:
Updated my plugin Version, Latest Release date and appropriate download link


jjw24:
Hi there, plugins will automatically update, no need to manually.

DiekoMA:
how long do they take to update ?

jjw24:
Should be every 3 hours, but we have recently discovered the CDN provider could be caching the manifest file so it may take up to 12 hours

PR #256, opened 2023-06-27
🔧 Update plugins.json

Galedrim:
Gitmoji plugin


Garulf (changes requested):
Please append plugin entry to the end of the list.

PR #259, opened 2023-07-31
Add adjustable Python version tests & allow testing of all Python plugins

jjw24:


jjw24:
My mistake, did not mean to merge this in without review.

PR #260, opened 2023-07-31
Add adjustable python ver test

jjw24:


jjw24:
related and same changes in #259

PR #270, opened 2023-08-25
Use asyncio to greatly improve performance

taooceros:
Given that our plugin number is not small, it makes sense to use asyncio to speed up the process.

- Workflow now takes 20s to finish, which are mostly setup time (compared to previouly around 1 miniute).
- Use `GITHUB_TOKEN` for authentication to avoid rate limit.


.github/workflows/updater.yaml
```diff
@@ -32,4 +32,4 @@ jobs:
         with:
           commit_message: "auto updated plugins"
           push_options: --force
-          branch: plugin_api_v2
+          branch: asyncio-revision
```
jjw24, line 35: This should be plugin_api_v2 instead?

taooceros: yes for testing purpose

taooceros:
close in favor of #319

PR #257, opened 2023-07-07
Add Bulk URL Opener

tarikjaber:

PR #278, opened 2023-09-04
Add Bulk URL Opener plugin

tarikjaber:
Previously closed here but wasn't sure how to repopen it: https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/pull/257


jjw24:
Could you have a look at the comments in the previous PR and see if they are resolved please.

tarikjaber:
What did you mean by "prepend the plugin name"?

I updated the way the plugin added the plugin folders to path using the method mentioned in the documentation here: https://www.flowlauncher.com/docs/#/py-setup-project?id=_3-use-lib-directory. What else could causing the plugin test run to fail?

plugins.json
```diff
@@ -1704,5 +1704,17 @@
...
+    },
+    {
+        "ID": "2d5036e2-eccc-4254-b88e-de78580e2195",
+        "Name": "Bulk URL Opener",
+        "Description": "Opens groups of tabs",
+        "Author": "Tarik Jaber",
+        "Version": "1.0.0",
+        "Language": "python",
+        "Website": "https://github.com/tarikjaber/Bulk-URL-Opener",
+        "UrlDownload":"https://github.com/tarikjaber/Bulk-URL-Opener/archive/refs/tags/v1.0.1.zip",
```
jjw24, line 1716: I mean instead of v1.0.1.zip, yse something like Bulk-URL-Opener.zip

Having had a look at the repo briefly, this can be actually linked with the latest release zip link:
https://github.com/tarikjaber/Bulk-URL-Opener/releases/download/v1.0.2/Flow.Launcher.Plugin.BulkURLOpener.zip

jjw24:
>What else could causing the plugin test run to fail?

The fail looks like unable to find your plugin's dependant packages. Are you able to download your plugin and run it with flow? If the same error, then check if the lib folder exists and the folder has been added to sys.path correctly.

PR #287, opened 2023-09-26
Add default plugins

jjw24:
Add all default plugins


onesounds (dismissed):

plugins.json
```diff
@@ -1739,19 +1739,161 @@
     },
     {
         "ID": "4f9e42bbcd9045e1969f97293dfaf4c7",
-        "ActionKeyword": "*",
```
VictoriousRaptor, line 1742: why deleting this line?

jjw24: Actionkeyword or executablefile keys are never needed for the manifest.

VictoriousRaptor (approved):

PR #288, opened 2023-09-27
Add plugins.json schema validation

VictoriousRaptor:


VictoriousRaptor:
Good POC but not ready. Output are not human-friendly. See here

jjw24:
Have a look at this python package, downside is validation is done using python script instead, but seems to work well- https://docs.python-cerberus.org/

PR #302, opened 2024-01-09
Add new plugin: Notion Search

MinYn:
Flow Launcher for Notion Search
[my repo](https://github.com/MinYn/flow-launcher-notion-plugin)

usage package lib
`flox-lib`, `notion-client`, `requests`

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -2004,5 +2004,17 @@
         "IcoPath": "https://cdn.jsdelivr.net/gh/Melv1no/Flow.Launcher.Plugin.easyssh/Flow.Launcher.Plugin.easyssh/Images/app.png",
         "DateAdded": "2024-01-04T00:32:56Z",
         "LatestReleaseDate": "2023-12-30T12:45:32Z"
+    },
+    {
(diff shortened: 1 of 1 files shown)


Garulf:
Looks like the plugin is unable to import Flox and crashing. This is most likely due to the _lib_ folder not being added to `sys.path`. 

The _test-run-plugin_ workflow attempts to simulate the embedded Python used by Flow Launcher. Unlike a normal C Python installation embedded Python does not add any folders to `sys.path` automatically.

It looks like you may be using a similar piece of code seen in my [Steam-Search](https://github.com/Garulf/Steam-Search) plugin. This unfortunately will not work the same as that plugin is compiled to an executable.

Please see a pure Python implementation like [HA-Commander](https://github.com/Garulf/HA-Commander) https://github.com/Garulf/HA-Commander/blob/main/main.py

MinYn:
@Garulf I modified the source by referring to the HA-Commander provided. Please check again.

Garulf (approved):
Thanks for your plugin submission! I know quite few users wanted this plugin.

PR #308, opened 2024-01-14
EasySSH update

Melv1no:
adding showmsg in case off ssh not installed / failing to create database


Garulf:
No need to post an update here. Any updates will be detected automatically.

PR #323, opened 2024-02-13
Fix plugin download url

jjw24:
Change to versioned download url so can be updated.


onesounds (approved):

jjw24:
Turns out autoupdater action actually fixes the download link in the notion plugin pr. No need for this one then.

PR #328, opened 2024-02-20
Plugin to trigger flow launcher by Win button

AminSallah:
**Key features:**

- Activate Flow Launcher using the Left Windows button as a substitute for Windows Search. 
- Maintain compatibility with LWin button shortcuts by holding the LWin key until Flow Launcher is hidden, and then use other button combinations.
- To access the Start menu, rapidly double-tap the Left Win key.
- Plugin effect is active only while Flow is running and the plugin is enabled.

**Considerations:**

- The plugin only affects the Left Windows button.
- The main Flow Hotkey must be set to Alt + Space.


jjw24 (approved):
💯 works really well. There is quite a high demand for the ability to use win key with flow so this is definitely a great plugin for those. Thank you for the submission.

PR #338, opened 2024-05-01
Add KeePass2Flow

geisterfurz007:
Adds KeePass2Flow, a new plugin that allows connecting to a KeePass2 database using the KeePassRPC plugin.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -2296,10 +2296,10 @@
         "Name": "KeePass2Flow",
         "Description": "Hook up Flow to KeePass 2",
         "Author": "geisterfurz007",
-        "Version": "1.0.0",
+        "Version": "1.1.0",
         "Language": "csharp",
         "Website": "https://github.com/geisterfurz007/KeePass2Flow",
-        "UrlDownload": "https://github.com/geisterfurz007/KeePass2Flow/releases/download/1.0.0/KeePass2Flow.zip",
+        "UrlDownload": "https://github.com/geisterfurz007/KeePass2Flow/releases/download/1.1.0/KeePass2Flow.zip",
         "UrlSourceCode": "https://github.com/geisterfurz007/KeePass2Flow",
         "IcoPath": "https://cdn.jsdelivr.net/gh/geisterfurz007/KeePass2Flow@master/Flow.Launcher.Plugin.KeePass2Flow/icon.png"
     }
--- a/plugins.json
+++ b/plugins.json
@@ -2296,10 +2296,10 @@
         "Name": "KeePass2Flow",
         "Description": "Hook up Flow to KeePass 2",
         "Author": "geisterfurz007",
-        "Version": "1.1.0",
+        "Version": "1.1.1",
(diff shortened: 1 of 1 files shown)


VictoriousRaptor:
Hi, I haven't had time to look into your plugin yet. But there was a KeePass plugin submission before [here](https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/pull/202) and we decided not to accept it because it's design was insecure. Please read the conversation in the PR I've posted and check if your plugin stores sensitive data in plain text.

geisterfurz007:
Hey, thanks for the heads-up @VictoriousRaptor ! I built the plugin to communicate using the KeePassRPC plugin, the same grounds of communication that the [Kee browser-addon](https://github.com/kee-org/browser-addon) uses. The only thing stored in plaintext is the key used to decrypt the communication between the RPC-Plugin and Flow-Plugin which is useless unless a database is opened. This also matches the behaviour of the Kee browser-addon which stores the key amongst a bunch of other config in `browser.storage.local` which is unencrypted as well.

To go in a bit more depth: The plugin I built fully relies on KeePass2 + KeePassRPC to access the database and thus requires the user to unlock their database through KeePass2. The plugin itself never knows the master password to any database it communicates with. The communication between my plugin and KeePassRPC is fully encrypted. The key is exchanged through an implementation of SRP which is pretty much a direct translation from the SRPc class of the browser-addon to C#. Once the key is exchanged and persisted, it is reused on a new connection after both the server and the client agree that the key is valid.

jjw24:
Hey thanks for making the change to store the key via Windows Credentials Manager. For sensitive plugins like this, this will have to be the requirement rather than persisting to disk so thanks for making the change.

jjw24:
@geisterfurz007 had a quick initial look and so far seems ok. I will take a look again later.

One thing I noticed is how the plugin is plubished to GitHub Releases. Could you please hook up Appveyor (if you are using it) to this project so the build and publish details are visible, otherwise GitHub Action Workflow is the other option.

geisterfurz007:
@jjw24 I can look into that yeah! You reckon using what the JetbrainsIDE Plugin uses should be enough?

https://github.com/kenty02/Flow.Launcher.Plugin.JetBrainsIDEProjects/blob/main/.github%2Fworkflows%2Fpublish.yaml

VictoriousRaptor:
> @jjw24 I can look into that yeah! You reckon using what the JetbrainsIDE Plugin uses should be enough?
> 
> https://github.com/kenty02/Flow.Launcher.Plugin.JetBrainsIDEProjects/blob/main/.github%2Fworkflows%2Fpublish.yaml

Yes, and many other plugins fyr.

geisterfurz007:
@VictoriousRaptor sweet! Yeah, Jetbrains was just the first to come to mind because I contributed the last update to it ^^ Added the GitHub Actions workflow and bumped the version to 1.1.1 both there and in this PR!

VictoriousRaptor (approved):

PR #345, opened 2024-05-13
✨ fix contextual menu Pokedex

Galedrim:


jjw24:
Perhaps just update the description only. Version will be updated automatically

PR #365, opened 2024-07-15
Add Git Easy plugin

zkwokleung:


taooceros:
I think there's a bug in which if the folder (e.g. the default `C:\Repos`) doesn't exist, the whole plugin wouldn't work. I think you may want to create the folder if not exist.

zkwokleung:
Thank you for your feedback. I have updated the plugin to handle such cases.
Results will be shown for the user to configure the path or create the folder.

jjw24 (approved):

jjw24:
@zkwokleung nice and creative icon design btw 👍

PR #383, opened 2024-08-26
Add new plugin: AppUpgrader

Exarilo:


taooceros:
emmm it doesn't seem to work with me? `up` just shows nothing?

Exarilo:
Oops, you're right, it should be fixed now.

jjw24 (approved):

PR #386, opened 2024-09-04
LinkOpener - UI improvement

Exarilo:


jjw24:
No need to submit version updates, it is automatically done.

PR #387, opened 2024-09-06
Added Open Window Search Plugin

Schumacher-Bastian:
Quickly switch to an open window.

**Features**

- Quickly search and switch between open windows
- Displays window titles and application names in search results
- Extracts and caches window icons for a visually rich experience
- Supports minimized and maximized windows


taooceros:
What's the difference between this and WindowsWalker?

PR #390, opened 2024-09-14
Add plugin: DirQuickJump

idkidknow:


taooceros:
interesting; does it support the open/save dialog?

@onesounds maybe you would be interested?

onesounds:
> interesting; does it support the open/save dialog?
> 
> @onesounds maybe you would be interested?

I haven't tested it. but It looks good if it works well. 😋

jjw24 (approved):
Tested open/save dialog of Edge + Adobe + Paint, working very nicely, thank you @idkidknow for making this 🥇 💯

PR #410, opened 2024-11-09
Add submission restriction list

jjw24:
Add submission restriction list


deefrawley (commented):
Suggest "Plugins that facilitate or contain any of the following will not be accepted, and while they may initially get uploaded to the Plugin Store via the automated CI process, they will be removed once discovered by the Flow team." What do you think?

jjw24:
Yes, they will be against the store policy and removed. I have updated the readme.

VictoriousRaptor (approved):

PR #413, opened 2024-11-15
change FirefoxKeywordBookmark's language from `python` to `python_v2`

cibere:
I recently updated my [`Firefox Keyword Bookmarks` plugin](https://github.com/cibere/FirefoxKeywordBookmarks) to [v0.1.0](https://github.com/cibere/FirefoxKeywordBookmarks/releases/tag/v0.1.0), which uses the V2 plugin api, so I changed the language key to `python_v2` in the repo, however the auto updater did not carry this update over to this repo.

tl;dr the auto updater didn't update my plugin's language key so I have to do it manually


taooceros:
technically it's fine because the only thing that matters is the language in `plugin.json` which will be updated when user update so no need to worry too much

taooceros (dismissed):

taooceros (approved):

PR #417, opened 2024-11-23
New Plugin Flow Load Notification

cibere:
Basically just adds a notification that gets shown when the plugin's initialize method is called

### DIFF
--- a/plugins/Flow Load Notification-5f48576d-efe7-4562-a4ed-fe64455f1ac1.json
+++ b/plugins/Flow Load Notification-5f48576d-efe7-4562-a4ed-fe64455f1ac1.json
@@ -1,0 +1,12 @@
+{
+  "ID": "5f48576d-efe7-4562-a4ed-fe64455f1ac1",  
+  "Name": "Flow Load Notification",
+  "Description": "A plugin that only sends a notification when the plugin gets loaded after a restart or on first start.",
+  "Author": "cibere",
+  "Version": "0.0.1",
+  "Language": "python_v2",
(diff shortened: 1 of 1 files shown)


jjw24:
Hey @cibere, what are the use cases for this plugin?

cibere:
> Hey @cibere, what are the use cases for this plugin?

I'm not sure about others, but I find it really annoying to have to guess when flow launcher is done loading or restarting, so i end up spamming my keybind until it shows up. I also have the flow launcher tray icon hidden because i use my tray icons a lot, and would find it even more annoying to either have an extra icon i won't use or have to drag flow's icon to the overflow menu.

This plugin doesn't need to be added to the manifest, I'm ok with it just being a local plugin for me, I just thought I'd create a PR anyways

jjw24:
>This plugin doesn't need to be added to the manifest, I'm ok with it just being a local plugin for me, I just thought I'd create a PR anyways

No it's fine, happy to add. I was asking because I wasn't sure if this is something we can fix on flow's side or this plugin is sufficient to fill that void.

Does it notifies only when the plugin finishes loading or when all plugins have finished?

cibere:
> Does it notifies only when the plugin finishes loading or when all plugins have finished?

It just notifies you when flow sends the plugin the initialize request, which for me at least, gives me a better idea as to when flow will be ready (in my testing, flow was normally ready 1-3 seconds after the notification depending on the other installed plugins

jjw24 (approved):

jjw24:
Oh ok so it notifies once this plugin's initialize has finished but there could be other plugins still initialising.

PR #430, opened 2024-12-13
Remove an incorrect duplicate plugin file

Yusyuriv:
I don't know much about the way the manifest works, but as far as I know, simply deleting the plugin file should be enough. Correct me if I'm wrong.

Seems like the original file was named incorrectly. It's called `Nightlight-95af3946-2bfe-49af-a9b6-edffa42ef1d5.json` but the id inside is `ef3a6ece-ca05-4d4f-92ba-2ff8a67e017b`. Update workflow then writes the new file with the correct name, but the old file is still there, causing infinite update notifications.


Garulf (approved):

Garulf:
Thank you.

PR #431, opened 2024-12-14
New Plugin: Unity projects launcher

Ghost-Miner:
plugin to open unity projects from the kauncher.

exiting plugins for this purpose no longer work and are abandoned or cancelled.


jjw24 (changes requested):
Could you please publish the release with GitHub action workflow.

Ghost-Miner:
How do you do that? There's no guide for C# and the file I made doesn't seem to work.

jjw24:
You are right, we are actually missing how to setup workflow for .Net plugins.

You almost got it, it's not triggering because the push rule is for master branch, your repo has main branch instead. So just need to change this line https://github.com/Ghost-Miner/Unity-project-launcher-plugin/blob/7ac9204725eef94b32071249d634cc2dd4963396/.github/workflows/Publish%20Release.yml#L5

Ghost-Miner:
so i uploaded a working workflow file, but now it **always** throws "path" not found and either doesnt tell me the full path at all or hides it buried somewhere

Ghost-Miner:
> so i uploaded a working workflow file, but now it **always** throws "path" not found and either doesnt tell me the full path at all or hides it buried somewhere

It looks like i _finally_ got the workflow working.

PR #432, opened 2024-12-15
Add filename construct validation

jjw24:
Test and validate plugin file name be in the construct of <name>-<id>.json.

Tested:
Previous Nightlight version fails validation

Relates to https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/pull/430


ci/src/_utils.py
```diff
@@ -53,7 +53,10 @@ def plugin_reader() -> P:
     return manifests
 
 def get_plugin_files() -> list[str]:
-    return [os.path.join(plugin_dir, file) for file in os.listdir(plugin_dir)]
+    return [os.path.join(plugin_dir, filename) for filename in get_plugin_filenames()]
+
+def get_plugin_filenames() -> list[str]:
+    return [file for file in os.listdir(plugin_dir)]
```
Yusyuriv, line 59: It looks like you don't need a list comprehension in `get_plugin_filenames` as they return the same data, but list comprehension does slightly more work as it iterates through list one more time:

```
>>> [v for v in os.listdir('test')]
['script2.py', 'plugin.json', 'file.txt', 'script.py']
>>> os.listdir('test')
['script2.py', 'plugin.json', 'file.txt', 'script.py']
```
```suggestion
    return os.listdir(plugin_dir)
```

Yusyuriv (approved):
LGTM. Removing the list comprehension is not very important as its impact on performance is negligible, so I approved the PR in case you don't want to remove the list comprehension.

ci/src/_utils.py
Yusyuriv: I didn't notice the PR was on auto-merge. Oh well.

jjw24: Ah yes, my mistake, will correct it in the next PR.

PR #425, opened 2024-12-04
Add Shortcut Plugin

mikegyi:
This PR adds the Shortcut plugin which allows Flow Launcher users to:
   - Search and create stories in Shortcut (formerly Clubhouse)
   - Plugin is built with TypeScript
   - Automated releases are set up via GitHub Actions

PR #435, opened 2024-12-17
Add plugin submission ID validation

jjw24:
Validates new submission plugin's ID is v4 UUID.

Only validate submission plugins and not for existing plugins because there are already invalid IDs which will not be possible to correct.

Resolves https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/pull/425#discussion_r1870533396


ci/src/validator.py
```diff
@@ -1,5 +1,9 @@
 # -*-coding: utf-8 -*-
-from _utils import clean, id_name, language_list, language_name, plugin_reader, check_url, icon_path, get_plugin_files, get_plugin_filenames
+import re
```
Yusyuriv, line 2: Is this import unused?

jjw24: Good catch, thank you. Removed.

jjw24:
```suggestion
```

ci/src/validator.py
```diff
@@ -35,3 +38,19 @@ def test_file_name_construct():
         assert (
             f"{info['Name']}-{info['ID']}.json" in filenames
         ), f"Plugin {info['Name']} with ID {info['ID']} does not have the correct filename. Make sure it's name + ID, i.e. {info['Name']}-{info['ID']}.json"
+
+def test_submitted_plugin_id_is_valid_uuid():
+    plugins_json_ids = [item["ID"] for item in get_file_plugins_json_info("ID")]
+    existing_plugin_file_ids = [info["ID"] for info in plugin_infos]
+
+    for id in existing_plugin_file_ids:
+        if id in plugins_json_ids:
```
jjw24, line 47:
```suggestion
        # plugins.json would not contain new submission's ID.
        if id in plugins_json_ids:
```

Yusyuriv (approved):

PR #436, opened 2024-12-17
Fixed incorrect icon url

Ghost-Miner:
Fixed invalid url for the Icon path which i haven't noticed before.


VictoriousRaptor:
Hi, please resolve conflicts so we can merge this pr. thx.

VictoriousRaptor (approved):

jjw24:
Hmm, though the validation check would have caught that error.

Ghost-Miner:
> Hmm, thought the validation check would have caught that error.

I think it would have to check if the HTTP response is 200. to catch this error.

PR #438, opened 2024-12-21
move screenbrightness plugin to plugins dir

cibere:
I accidently put my ScreenBrightness plugin from #434 in the root dir, not the plugin dir.  This PR just moves it to the right dir


jjw24 (approved):
My mistake, I didn't catch that either.

PR #440, opened 2024-12-29
Re-add Google Search plugin

jjw24:
Re-add after fix. Following on from #264


Yusyuriv:
Wouldn't the download link need to be to v2.0.1? Or does it get automatically updated for all new plugins?

deefrawley (approved):

PR #444, opened 2025-01-13
Add Seach Unicode plugin set

blueset:
Repository and README: https://github.com/blueset/flow-search-unicode

### DIFF
--- a/plugins/SearchUnicode.Emoji-BB55C11CB9FB4AAF81B670D9FF0BFBEE.json
+++ b/plugins/SearchUnicode.Emoji-BB55C11CB9FB4AAF81B670D9FF0BFBEE.json
@@ -4,11 +4,11 @@
     "Name": "Search Unicode – Emoji",
     "Description": "Search Emoji in the Unicode character set",
     "Author": "Eana Hufwe",
-    "Version": "1.0.0",
+    "Version": "1.0.1",
     "Language": "csharp",
     "Website": "https://github.com/blueset/flow-search-unicode",
     "ExecuteFileName": "Flow.Launcher.Plugin.SearchUnicode.Emoji.dll",
     "IcoPath": "https://cdn.jsdelivr.net/gh/blueset/flow-search-unicode@master/Flow.Launcher.Plugin.SearchUnicode.Emoji/icon.png?raw=true",
     "UrlSourceCode": "https://github.com/blueset/flow-search-unicode",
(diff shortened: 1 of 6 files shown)


jjw24:
Hi there, how's it going?

Could you please confirm where the uni.exe file is downloaded from https://github.com/blueset/flow-search-unicode/blob/master/uni.exe. Could you please use the GitHub workflow or the release.ps1to down this file instead, so that others can confirm the source of this exe.

Let me know if you need further assistance with this :)

blueset:
Updated the repo and pipeline to retrieve `uni.exe` at build time instead of bundling in git repo.

Please find the download script here: https://github.com/blueset/flow-search-unicode/blob/master/get-uni.ps1

plugins/SearchUnicode.Emoji-BB55C11CB9FB4AAF81B670D9FF0BFBEE.json
```diff
@@ -0,0 +1,14 @@
+{
+    "ID": "BB55C11CB9FB4AAF81B670D9FF0BFBEE",
+    "ActionKeyword": "e",
```
jjw24, line 3:
```suggestion
```

jjw24:
Awesome, thanks for fixing that up. I forgot to mention if you ever wanted to actually handle it in your code instead of a standalone script so that when you develop locally the file gets downloaded as well without having to separately run the script, you can have a look at my Droplex library. The [Explorer plugin](https://github.com/Flow-Launcher/Flow.Launcher/blob/0b8db59e39e71ccd305f8b240dbbdf8269e3a0d3/Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingDownloadHelper.cs#L47) as an example uses it to download and extract the Everything service.

jjw24 (approved):

blueset:
Thanks for the suggestion! If I understand correctly, Droplex is like a combination of a subset of a package manager and software repository that exposes a C# API. In the case of my plugins, since `uni` distributes as a portable binary, the installation process would simply be download and unzip. Ideally I hope to hand over binary dependencies to `winget` which is shipped with Windows 11 to avoid having multiple copies of the same binary in file system, but currently its community repository doesn’t allow portable software lacking a binary installer.

jjw24:
>If I understand correctly, Droplex is like a combination of a subset of a package manager and software repository that exposes a C# API.

Yes, pretty much a package manager that lives with the application code and exposes an API for easily managing silent installs and zip extraction. The goal of Droplex is to enable the application to install its required third party services independent of a package manager.

Issue #420, opened 2024-11-28
[FEATURE] New Plugin Discord Webhook Notification


cibere:
In the flow discord server, there is a webhook for when plugins get updated, but what about adding a message to that channel when a new plugin gets accepted?

cibere:
> Yeah sounds good.
> 
> Something like this:
> - Can add the code here to post- https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/blob/plugin_api_v2/ci/src/discord.py
> - Maybe move this [send notification function](https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/blob/2c5ff3ee31105bc5a2be62ded727f1d932a06063/ci/src/updater.py#L92) into discord.py
> - New ci to trigger on merge. Figure out how to trigger only for plugin submissions, maybe in the same time the `DateAdded` field is added to the plugin file.
> 

I applied all three of those changes in #453

cibere:
Closing since #453 has been merged

PR #453, opened 2025-01-29
add plugin release webhook notification

cibere:
Based on #420 

This PR adds a notification that gets triggered in the `python-passed.yml` workflow, where when the `DateAdded` key gets added to a plugin, a notification is sent to discord stating that a new plugin has been uploaded. I tried to keep the embed similar to the one used for updates, but I made a couple of changes.

## Notification Example
image

## Overview of changes
1. Added requirements to the `update-tested.py` script and `python-passed.yml` workflow
2. added a `release_hook` function based on `update_hook` to `discord.py`
3. have `update-tested.py` execute `release_hook` for new releases
4. Added webhook_url to `update-tested.py` via a CLI argument 
5. I also fixed a couple of typehints so that my typechecker would stop yelling at me

### DIFF
--- a/ci/src/discord.py
+++ b/ci/src/discord.py
@@ -44,15 +44,15 @@ async def update_hook(webhook_url: str, info: PluginType, latest_ver: str, relea
     async with aiohttp.ClientSession() as session:
         await session.post(webhook_url, json=embed)
 
-async def release_hook(webhook_url: str, info: PluginType, latest_ver: str, release: dict) -> None:
+async def release_hook(webhook_url: str, info: PluginType) -> None:
     embed = {
         "content": None,
         "embeds": [
             {
             "title": info[plugin_name],
-            "description": f"New Plugin!\nReleased at v{latest_ver}.",
-            "url": release['html_url'],
-            "color": 5763719,
+            "description": f"New Plugin!\nReleased at v{info[version]}.",
+            "url": info[website],
+            "color": 5763719, # green
             "fields": [
                 {
                 "name": "Plugin Description",
@@ -77,9 +77,7 @@ async def release_hook(webhook_url: str, info: PluginType, latest_ver: str, rele
         embed['embeds'][0]['author']['name'] = github_username
         embed['embeds'][0]['author']['url'] = f"{github_url}/{github_username}"
         embed['embeds'][0]["author"]["icon_url"] = f"{github_url}/{github_username}.png?size=40"
-    release_notes = release.get('body')
-    if release_notes and release_notes.strip():
-        embed['embeds'][0]['fields'].append({"name": "Release Notes", "value": truncate_release_notes(release['html_url'], release.get('body', ""))})
+    
     async with aiohttp.ClientSession() as session:
         await session.post(webhook_url, json=embed)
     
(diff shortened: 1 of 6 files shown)


ci/src/discord.py
```diff
@@ -59,3 +98,17 @@ def truncate_release_notes(url: str, release_notes: str, length: int = MAX_BODY_
...
     return release_notes[:graceful_truncation_index] + TRUNCATION_MESSAGE
+
+async def send_notification(
+    info: PluginType, latest_ver, release, webhook_url: str | None = None, is_release: bool = False
+) -> None:
+    if not webhook_url:
+        return
+    
+    if is_release or version_tuple(info[version]) != version_tuple(latest_ver):
+        tqdm.write(f"Update detected: {info[plugin_name]} {latest_ver}")
```
jjw24, line 109: Maybe this should be 'Sending Discord notification'

ci/src/update-tested.py
```diff
@@ -2,23 +2,38 @@
...
+    etags = etag_reader()
 
     for idx, plugin in enumerate(plugin_infos):
         if plugin["Language"] == "python" and "Tested" not in plugin.keys():
             plugin_infos[idx]["Tested"] = True
         # Add date added if field is not present
         if plugin.get(date_added) is None:
-            plugin_infos[idx][date_added] = datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ")
+            plugin_infos[idx][date_added] = datetime.now(UTC).strftime("%Y-%m-%dT%H:%M:%SZ")
+            yield batch_github_plugin_info(plugin, etags, github_token, webhook_url, True)
```
jjw24, line 28: Can we just call the discord send notification here instead of doing this? There won't be many new plugins at once.

cibere: I suppose, however the webhooks wouldn't include as much data. for example: the release notes

cibere: > There won't be many new plugins at once.

Ah i think i originally misread what you said. Yeah, I can move away from using asyncio.gather

jjw24: I was ultimately thinking to keep this call simpler. Without batch_github_plugin_info what else do we miss out on, is it just release notes? I think for a new plugin notification there isn't much need for release notes, most the  information should be able to be found in the plugin file like title, author and description right?

cibere: > I was ultimately thinking to keep this call simpler. Without batch_github_plugin_info what else do we miss out on, is it just release notes? I think for a new plugin notification there isn't much need for release notes, most the information should be able to be found in the plugin file like title, author and description right?

After a little bit of reworking, I was able to replace everything that came from the github request except the release notes.

image

cibere:
I also reverted most of the changes I made to `send_notification` and opted to interact directly with `release_hook`, since `release_hook` and `send_notification` take different parameters now.

cibere:
I decided to move `send_notification` back to `updater.py` since it's no longer being used in `update-tested.py`, and there are 2 discord webhook hooks in `discord.py`

jjw24 (approved):
👍

PR #458, opened 2025-02-04
Add VirusTotal scan action workflow

jjw24:
Part of https://github.com/Flow-Launcher/Flow.Launcher/discussions/3121


cibere:
Well... this would fail when testing my pyrepl plugin 😓

jjw24:
Hmm, not sure what you mean, this would not have any impact on testing. It won't be run as part of testing.

cibere:
> Hmm, not sure what you mean, this would not have any impact on testing. It won't be run as part of testing.

The intention is to run the plugin's extracted files through virustotal, right? If you do that with my pyrepl plugin, it will be flagged due to the standalone exe

jjw24:
>The intention is to run the plugin's extracted files through virustotal, right? 

Yes.

>If you do that with my pyrepl plugin, it will be flagged due to the standalone exe

Is it flagging because it's standalone exe or something else though?

cibere:
> > If you do that with my pyrepl plugin, it will be flagged due to the standalone exe
> 
> Is it flagging because it's standalone exe or something else though?

It's because of the way pyinstaller makes the standalone exe. My understanding is that it uses methods similar to trojans, so virustotal flags it. Pyinstaller doesn't have to build it as a standalone exe though, that was just the choice I made when making the build workflow.

jjw24:
Superseded by #694

Issue #459, opened 2025-02-05
[BUG] wrong download url for plugins that share a repo


cibere:
I recently opened https://github.com/Flow-Launcher/Flow.Launcher/issues/3226, however after investigating, I realized that the problem was with the auto updater.

Specifically [line 49 in `updater.py`](https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/blob/5d00f859e8577ef918fdaa6884ced41abcdec902/ci/src/updater.py#L49):
```py
info[url_download] = assets[0]["browser_download_url"]
```

The updater grabs the first release artifact, which works for plugins that have their own repo, but doesn't work in the case of the unicode search plugin set.

I'm happy to PR a fix, I'm just not sure what fix you guys would prefer.

## Potential Solution
One potential solution would be to use the existing `DownloadUrl` value to get the name of the zip file, then attempt to find a zip file with the same name in the release's artifacts. If no file was found, grab the first one.

This would keep the current behavior for plugins that have their own repo, and if plugins share a repo, would require the plugin dev to ensure that the filename of the artifact in the release matches the expected filename in the plugin's manifest entry.

## Affected Plugins
As of now, the only plugins that seem to be impossible to download through the manifest are the `Search Unicode - Identity` and `Search Unicode - Search` plugins

jjw24:
Good find. Solution sounds good.  Filename of the artifact in the release must match the filename in the plugin's manifest entry.

PR #460, opened 2025-02-05
Update Search Unicode Search and Identify Download URLs

jeregon1:
These 2 plugins download URLs led to Unicode Search Emoji instead, as you can see.


cibere:
Unfortunately the solution isn't that simple. The auto updater is grabbing the wrong download url, so the auto updater needs to be modified, not the plugin entries. See #459 for more info

jjw24:
This is now fixed thanks to cibere.

PR #461, opened 2025-02-05
Update "Search Unicode – Identify" Download URL

jeregon1:
The original led to Search Unicode Emoji plugin download URL


cibere:
Unfortunately the solution isn't that simple. The auto updater is grabbing the wrong download url, so the auto updater needs to be modified, not the plugin entries. See #459 for more info

PR #462, opened 2025-02-05
Fix wrong download urls with shared repos

### LINKED ISSUE #459: [BUG] wrong download url for plugins that share a repo

cibere:
This PR fixes #459 using the potential solution I mentioned there.

## Quick Overview
Instead of fetching the first release artifact, check the amount of assets, and if there are more than 1, attempt to find one with the same name as the current download url. If there is only 1 asset, or no asset with the same filename could be found, then default to the first asset.

This would leave the current behavior for plugins that have their own repository, but for plugins that share a repository, would require the plugin dev to ensure the proper artifact names correspond to the correct plugin.

When it comes to getting the original filename, I decided to use a regex:
```py
https://github\.com/(?P<author>[a-zA-Z0-9-]+)/(?P<repo>[a-zA-Z0-9\.\-\_]*)/releases/download/(?P<version>[a-zA-Z\.0-9]+)/(?P<filename>.*)\.zip
```
which I've tested against the current set of plugins, and made it fit the username/repo name requirements that I could find.

If you want me to move away from the regex, to something like this:
```py
filename = info[url_download].split("/")[-1]
```
then lmk.

I also went ahead and fixed the urls in the `Search Unicode - Identify` and `Search Unicode - Search` plugins so that another PR isn't needed to do that


cibere:
Not exactly sure why the link checker workflow is failing. It says a configuration error, but it only seems to fail on this PR

jjw24 (approved):

PR #465, opened 2025-02-09
add UvxPy plugin

cibere:
https://github.com/cibere/Flow.Laucher.Plugin.uvxpy


cibere:
wow I really messed up those links 🤦

PR #470, opened 2025-02-23
New Plugin: Wallpaper Changer

ImNotVarun:
Change Wallpaper with the help of Pexel Api


cibere:
Is there a specific reason that you've chosen to not use the builtin settings system?

ImNotVarun:
I used a text file for storing the API key to keep it simple. If using the built-in settings system is preferred, I can update it

cibere:
> I used a text file for storing the API key to keep it simple. If using the built-in settings system is preferred, I can update it

Plugin settings are stored in a different place than the plugins themselves, so could you please either store your file in that location or switch to the builtin settings system?

ImNotVarun:
C:\Users\Username\AppData\Roaming\FlowLauncher\Settings\Plugins\Wallpaper
Changer
i am storing the file in this location is that correct

cibere:
> C:\Users\Username\AppData\Roaming\FlowLauncher\Settings\Plugins\Wallpaper Changer i am storing the file in this location is that correct

That seems to be the correct directory, however I still think it would be best to use the [builtin settings system](https://www.flowlauncher.com/docs/#/json-rpc-settings).

In case you are unaware how to use the builtin settings system, see the [docs](https://www.flowlauncher.com/docs/#/json-rpc-settings) for how to define the settings, and for getting them in your code, iirc you can use `self.rpc_request["settings"]` in your `query` method to get a dict containing the settings that have been sent with the request.

ImNotVarun:
I've updated the plugin to use the built-in settings system as suggested using flox. The API key is now stored within Flow Launcher's settings system instead of a text file

cibere:
I appreciate you switching your storage method to use flow's settings system, however I also wanted you to use the [`SettingsTemplate.yaml`](https://www.flowlauncher.com/docs/#/json-rpc-settings?id=settingstemplateyaml) file that lets you customize the plugin's settings interface, so that it aligns with how the other plugins work.

You can keep your current code, I'm just asking for the addition of that file

ImNotVarun:
Done , added the SettingsTemplate.yaml file

ImNotVarun:
Hey @cibere, I’ve made all the changes you suggested ,  Let me know if there’s anything else needed

cibere:
> Hey @cibere, I’ve made all the changes you suggested , Let me know if there’s anything else needed

Thank you for all of the changes that you've made.

PR #479, opened 2025-03-07
Change Clipboard+ description

Jack251970:


cibere (approved):
btw `plugins.json` is auto generated from the plugin files in the `plugins` dir

Jack251970:
> btw `plugins.json` is auto generated from the plugin files in the `plugins` dir

Will it break something?

cibere:
> > btw `plugins.json` is auto generated from the plugin files in the `plugins` dir
> 
> Will it break something?

Only if you break the formatting (that file is what flow uses to find plugins)

PR #480, opened 2025-03-07
Sync plugin metadata info fields during plugin update

cibere:
When an update is detected in the auto updater, the manifest's copy of the plugin's metadata doesn't get its information keys synced with the plugin's `plugin.json` file, leading to outdated descriptions, urls, authors, etc.
*see all of the plugin files that I updated in this PR because they had out-of-date info*

This PR adds code that fetches the plugin's `plugin.json` file when an update is detected, and updates the following fields:
- `Description`
- `Author`
- `Language`
- `Website`

Note: Two plugins will error with this due to a typo with their `Description` keys. I've created PRs on their repos to fix this ([CursorWorkspace](https://github.com/cou723/Flow.Plugin.CursorWorkspace/pull/1) and [VSCodeWorkspace](https://github.com/taooceros/Flow.Plugin.VSCodeWorkspace/pull/18))


plugins/Fancy Emoji-FF2C31D6A47348FFB9ED4EB26F6794E2.json
```diff
@@ -1,11 +1,11 @@
 {
     "ID": "FF2C31D6A47348FFB9ED4EB26F6794E2",
     "Name": "Fancy Emoji",
-    "Description": "Search for emoji and add to your clipboard automatically",
+    "Description": "Search for emoji and add to your clipboard automatically.",
     "Author": "Mave",
     "Version": "1.0.9",
     "Language": "python",
-    "Website": "https://github.com/Ma-ve/Flow.Launcher.Plugin.FancyEmoji",
+    "Website": "https://github.com/Ma-ve",
```
jjw24, line 8: This doesn't seem to be correct

ci/src/updater.py
```diff
@@ -68,14 +70,28 @@ async def batch_github_plugin_info(
                     browser_download_url or assets[0]["browser_download_url"]
                 )
 
-                if webhook_url:
-                    await send_notification(
-                        info,
-                        clean(latest_rel["tag_name"], "v"),
-                        latest_rel,
-                        webhook_url,
+                info[version] = latest_ver = clean(latest_rel["tag_name"], "v")
```
jjw24, line 73: Is this correct?

jjw24:
This is fine for now since we don't get too many plugin updates per run. We might need to revisit this in the future to separate the metadata updates out on its own and implement the same etag checking (if possible) to avoid doubling up the GitHub API requests during updates.

cibere:
> This is fine for now since we don't get too many plugin updates per run. We might need to revisit this in the future to separate the metadata updates out on its own and implement the same etag checking (if possible) to avoid doubling up the GitHub API requests during updates.

I don't understand how that would help, since the second request to the github api only happens when an update is detected, my understanding is that the current etag system will stop the function if a new update isn't detected before it reaches the new code, and even if it doesn't, there is a version check to ensure that its a new update

ci/src/updater.py
cibere: > Is this correct?

In that line I'm setting it to `info[version]` for the `plugin.json` file, and to the `latest_ver` variable so that I can use it later on without using the `info` dict

If you're asking if its a valid python statement, then yes. You can set values to multiple variables in a single statement. ex: `a = b = c = 0`, which sets, `a`, `b`, and `c` to `0`.

jjw24: Yeah but you are comparing info[version] and latest_ver in the next statement?

jjw24:
Using etag will send a conditional request, which doesn't count against the API limit. So updating metadata based on etag is ideal in the future when plugin updates occur a lot more, and for right now since there aren't many instances where plugin actually is updated in one run it's ok to send an API request for release info and another to download the zip file.

ci/src/updater.py
cibere: Ah yeah. Thanks for noticing. That comparison is how the send notification function worked, and I just copy pasted the statement. Guess I need to compare the clean version to the unclean version, then update the key.

cibere:
> Using etag will send a conditional request, which doesn't count against the API limit if content is the same (it returns 304 instead of retrieving the requested result). So updating metadata based on etag is ideal in the future when plugin updates occur a lot more, and for right now since there aren't many instances where plugin actually is updated in one run it's ok to send an API request for release info and another to download the zip file.

Ah ok, I think I understand now.

Anyways, I believe I've fixed the problem you mentioned in your review

jjw24:
This function name `batch_github_plugin_info` really confuses me- is it the current plugin info or new updated info? Everytime I read this I need to think twice what context this is in... Next time if you happen to change this part of the code please do me a favour and update this name to something better describe this functionality e.g. `get_updated_batch_plugin_info` 🙏

PR #484, opened 2025-03-14
New Plugin: Lively Wallpaper Controller

Odotocodot:
Source: https://github.com/Odotocodot/Lively4Flow
Preview:
gif


jjw24:
Just a suggestion, I think instead of hooking into visibility change event, which triggers everytime flow is shown, it might also work just via the query method. If you hook into the event when users each time trigger flow to search then your code to look for the Lively process is ran as well. 

Alot of times it's better to narrow down the plugin processing to when the plugin is actually called/used. I don't think the get process call is a heavy one so I will approve this.

jjw24 (approved):

Odotocodot:
> Just a suggestion, I think instead of hooking into visibility change event, which triggers everytime flow is shown, it might also work just via the query method. If you hook into the event when users each time trigger flow to search then your code to look for the Lively process is ran as well.

Thanks for the heads up. I've moved it to the `Query` method, with a boolean set in the `VisbilityChanged` event.
Also, thanks for the merge 😃.

PR #485, opened 2025-03-19
New Plugin: Text Formatter

Koalhack:
Source: https://github.com/Koalhack/Flow.Launcher.Plugin.Formatter
Preview: https://github.com/user-attachments/assets/8f09e6cb-43d0-4595-bac1-79d079f7a0f2

### DIFF
--- a/plugins/Text Formatter-3cd1d9a8e92e4210a0fd30d1d65bb664.json
+++ b/plugins/Text Formatter-3cd1d9a8e92e4210a0fd30d1d65bb664.json
@@ -1,0 +1,12 @@
+{
+    "ID": "3cd1d9a8e92e4210a0fd30d1d65bb664",
+    "Name": "Text Formatter",
+    "Description": "Retrieve text from the clipboard and edit it using several custom scripts",
+    "Author": "Koalhack",
+    "Version": "0.0.1",
+    "Language": "javascript",
+    "Website": "https://github.com/Koalhack/Flow.Launcher.Plugin.Formatter",
+    "UrlDownload": "https://github.com/Koalhack/Flow.Launcher.Plugin.Formatter/releases/download/v0.0.1/Flow.Launcher.Plugin.Formatter.zip",
+    "UrlSourceCode": "https://github.com/Koalhack/Flow.Launcher.Plugin.Formatter",
+    "IcoPath": "https://cdn.jsdelivr.net/gh/Koalhack/Flow.Launcher.Plugin.Formatter@main/icon/app.png",
+}


plugins/Text Formatter-3cd1d9a8e92e4210a0fd30d1d65bb664.json
```diff
@@ -0,0 +1,12 @@
...
+    "ID": "3cd1d9a8e92e4210a0fd30d1d65bb664",
+    "Name": "Text Formatter",
+    "Description": "Retrieve text from the clipboard and edit it using several custom scripts",
+    "Author": "Koalhack",
+    "Version": "0.0.1",
+    "Language": "javascript",
+    "Website": "https://github.com/Koalhack/Flow.Launcher.Plugin.Formatter",
+    "UrlDownload": "https://github.com/Koalhack/Flow.Launcher.Plugin.Formatter/releases/download/v0.0.1/Flow.Launcher.Plugin.Formatter.zip",
+    "UrlSourceCode": "https://github.com/Koalhack/Flow.Launcher.Plugin.Formatter",
+    "IcoPath": "https://cdn.jsdelivr.net/gh/Koalhack/Flow.Launcher.Plugin.Formatter@main/icon/app.png",
```
cibere, line 11:
```suggestion
"IcoPath": "https://cdn.jsdelivr.net/gh/Koalhack/Flow.Launcher.Plugin.Formatter@main/icon/app.png"
```
workflows are failing due to this trailing comma

cibere (changes requested):

Koalhack:
Hi @cibere, 

This is indeed an explicit design decision, because in future updates I'll be adding the ability for users to create their own custom scripts and if the user encounters an error, Flow Launcher's error system makes it easier to read the exception and help the user debug their script.

And yes, the description of my plugin at this stage of development is not accurate, it will do this kind of thing in the future. I've modified the description for now, for a better description.

```json
"Description": "Retrieve text from the clipboard and edit it using several custom scripts"
```

cibere:
> Hi @cibere,
> 
> This is indeed an explicit design decision, because in future updates I'll be adding the ability for users to create their own custom scripts and if the user encounters an error, Flow Launcher's error system makes it easier to read the exception and help the user debug their script.

I see. Though as far as I could tell, the errors were being ignored and not sent to flow's error system. The only sort of logs that I can find regarding it are in flow's log file itself, which just says:
```
11:51:03.5303-07:00 - WARN  - JsonRPCPlugin.ExecuteAsync - Empty JSON-RPC Response.
```

> And yes, the description of my plugin at this stage of development is not accurate, it will do this kind of thing in the future. I've modified the description for now, for a better description.

thank you for editing it. Once this plugin is done, and custom script support is added, I look forward to how useful it will be :)

Koalhack:
Hi @cibere,

Thanks for your reply, I'll be updating my plugin for custom scripts soon and adding more built-in scripts.

By the way, I don't know if I've understood what errors you're referring to, are you talking about script loading errors (metas, name, etc.) and script execution errors (code, context) or are we talking about something else?

Then I'd like to manage the errors as well as possible for future users but according to the documentation my only options would be: 
- Display the error message directly in the Flow Launcher Query results. ()
- Send a message using the Windows notification system `Flow.Launcher.ShowMsg`

I'm not against these solutions, but would there be a way like the error popup displayed by Flow Launcher?

I thank you in advance for your answer and for your participation in maintaining this beautiful tool that is Flow Launcher.

cibere:
> By the way, I don't know if I've understood what errors you're referring to, are you talking about script loading errors (metas, name, etc.) and script execution errors (code, context) or are we talking about something else?

I believe they are script execution errors, however I'm not sure about anything else since the plugin is suppressing them.

> Then I'd like to manage the errors as well as possible for future users but according to the documentation my only options would be:
> 
>     * Display the error message directly in the Flow Launcher Query results. ()
> 
>     * Send a message using the Windows notification system `Flow.Launcher.ShowMsg`
> 
> 
> I'm not against these solutions, but would there be a way like the error popup displayed by Flow Launcher?

With JSON-RPC plugins, flow will display the error message if a plugin raises/throws an error that never got handled (thus why I said that your plugin must be suppressing them). I'm not very familiar with js/ts, so perhaps @Yusyuriv can help with identifying how the errors are getting supressed.

Koalhack:
> With JSON-RPC plugins, flow will display the error message if a plugin raises/throws an error that never got handled (thus why I said that your plugin must be suppressing them). I'm not very familiar with js/ts, so perhaps @Yusyuriv can help with identifying how the errors are getting supressed.

Thank you for your fast reply. Could you tell me what modification you made to one of my scripts to get an error, or is it an error without having done anything?

Yusyuriv:
Sorry, I'm not very familiar with they way JSON-RPC works, but don't you suppress all script errors here?

```ts
  run(context: ScriptExecution) {
    vm.createContext(context);
    try {
      this.vmScript.runInContext(context);
    } catch (e) {
      console.error(`Error in VM Script execution: ${e}`);
    }
  }
```

What if you remove this try..catch?

Koalhack:
Hi @Yusyuriv, 

The `try...catch()` does not suppress the error, it is supposed to "catch" all errors and currently this portion of code only executes if there is an error with the context (the clipboard value) and so far it has never been triggered yet it is this `try... catch()` which is executed in the case of a script writing error.

```typescript
try {
      let script: string = fs.readFileSync(path).toString();

      let meta = stringRange(script, metaStartTerm, metaEndTerm);
      let jsonMeta = JSON.parse(meta) as Metas;

      let scriptObject = new Script({
        script: script,
        parameters: jsonMeta,
        builtIn: true
      });

      this.scripts.push(scriptObject);
    } catch (e) {
      console.error(`Unable to load ${path}, Error: ${e}`);
    }
  }
```

But as I said before, it will just display the error as a `print` in python, but as there's no error display system in Flow Launcher (JsonRPC), I think I'll display the errors in the query results.

cibere:
> but as there's no error display system in Flow Launcher (JsonRPC)

There is, and you can see that yourself by throwing an error at the start of your program in your main file. something in your code is suppressing the errors that get raised elsewhere in your code.

> I think I'll display the errors in the query results.

Alright. let me know when you are finished

Koalhack:
Hi,

I've updated my plugin for better error handling + support for user scripts.

(sorry for the delay I've been quite busy)

cibere:
Thank you for the work that you've put into this plugin

cibere (approved):

PR #492, opened 2025-04-03
Remove Periodic Table

Lucas256-cmd:
I would like to remove my plugin, I know this is not commonly done but there was a better plugin already. Called Element Flow and I think it is better to contribute and use to that plugin.


cibere:
The `plugins.json` file is automatically generated from the files in the `plugins` directory, so in order for your plugin to be removed, you need to delete its file from the `plugins` directory.

Lucas256-cmd:
Thanks, should be done now

cibere (approved):

PR #494, opened 2025-04-23
JWT Token Decoder plugin

Crakzzy:
JWT Decoder plugin, that decodes provided JWT token and also splits it into parts that user can copy into clipboard.
image


cibere (changes requested):
This is not a json file

Crakzzy:
> This is not a json file

Oh, I forgot to put .json to the filename. It should be good now @cibere

plugins/JWTDecoder-d2389086abe1495a8b476f39aa5c7b6f.json
```diff
@@ -0,0 +1,12 @@
+{
+  "ID": "d2389086abe1495a8b476f39aa5c7b6f",
+  "Name": "JWT Decoder",
```
cibere, line 3: Plugin name here doesn't match the plugin name in the filename

cibere (changes requested):

plugins/JWTDecoder-d2389086abe1495a8b476f39aa5c7b6f.json
Crakzzy: It should be fixed now

cibere (approved):
I hope you don't mind, but I renamed the plugin back to `JWT Decoder` since that was the original name, and the name in the `plugin.json` file, and since I assume you removed the space as to not have a space in the filename

Crakzzy:
> I hope you don't mind, but I renamed the plugin back to `JWT Decoder` since that was the original name, and the name in the `plugin.json` file, and since I assume you removed the space as to not have a space in the filename

No I don't mind at all. Thanks for your patience with me and I'm glad I could contribute 😄

PR #498, opened 2025-04-28
Validate download url in `validator.py`

cibere:
I've noticed quite a few plugin submissions with the download url set to the releases page of their github repo, so this PR simply adds a test to `validator.py` that validates the download url using the download url regex introduced in #462.


jjw24 (approved):

jjw24:
👍

PR #502, opened 2025-05-16
New Plugin: DayNight Toggle

jatinderjit:
A plugin to switch between Dark mode and Light mode.

### DIFF
--- a/plugins/DayNight Toggle-03614b3c-9e7d-4bc5-9509-d7f556f1a2b1.json
+++ b/plugins/DayNight Toggle-03614b3c-9e7d-4bc5-9509-d7f556f1a2b1.json
@@ -1,0 +1,12 @@
+{
+  "ID": "03614b3c-9e7d-4bc5-9509-d7f556f1a2b1",
+  "Name": "DayNight Toggle",
+  "Description": "Switch between Light and Dark mode",
+  "Author": "Jatinderjit Singh",
+  "Version": "1.0.3",
+  "Language": "python",
+  "Website": "https://github.com/jatinderjit/Flow.Launcher.Plugin.DayNightToggle",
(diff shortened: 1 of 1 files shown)


cibere:
Hello, and thank you for this plugin submission. Everything looks good to me, however is there is a specific reason for all of the default keywords? It seems a little excessive to me, especially since they all seem to do the same thing, and there is no way to either edit the keywords, or remove any that conflict with other plugins

jatinderjit:
Hi. Thanks for reviewing this.
I don't know what the standard practice is for the keywords. I just intended to make the search easy, so that there's not too much cognitive load to remember what the keyword is.

What I wanted was to start showing the option while the user is typing. For example, the results should appear when the user has typed `lig`. But I couldn't find a way as Flow seems to be keyword-oriented. So added other shortcuts like "dn", "th".

cibere:
> Hi. Thanks for reviewing this.
> 
> I don't know what the standard practice is for the keywords. I just intended to make the search easy, so that there's not too much cognitive load to remember what the keyword is.
> 
> 
> 
> What I wanted was to start showing the option while the user is typing. For example, the results should appear when the user has typed `lig`. But I couldn't find a way as Flow seems to be keyword-oriented. So I added other shortcuts like "dn", "th".
> 
> 
> 
> Do you have any suggestions for me? Does one of these seem fine?
> 
> 1. `day`, `night`, `dn`, `dark`, `light`
> 
> 2. `dark`, `light`
> 
> 3. Just `dn`
> 
> 4. something else?

With the latest version of flow, if you only have one keyword, flow will make it easy for people to change it. However, having more than one registered makes flow hide this customization, and requires the plugin dev to implement it themselves. With that said, unless you either have a good reason for multiple, or you implement a way to edit the keywords, I'd like you to switch to a single keyword. Perhaps `th` or `sth` (for system theme)

jatinderjit:
> implement a way to edit the keywords

I didn't know this was possible! I'll check that out.

Meanwhile, I've updated the plugin to use only one keyword `dn`.

cibere (approved):

PR #513, opened 2025-06-04
Add Window Manager plugin

Jack251970:
Since I have just set this repository to public, we need to wait for a while to make `IcoPath` able to work.


cibere (approved):

Jack251970:
> This plugin doesn't appear to work for me. I see the results, however nothing happens upon clicking on them, and I see no evidence of any errors being logged anywhere

Can you sure the foreground window is not a elevated one?

PR #516, opened 2025-06-05
New Plugin: FileFolderOpener

Exarilo:
This plugin allows users to open files or folders using keywords set in the settings. It should behave the same way as my other plugin LinkOpener


cibere (approved):
Looks good to me, however I didn't manually test it with flow

PR #526, opened 2025-06-25
Create Env-0F43C966617C4817B1BEC29DCC8E5A69.json

lurebat:


cibere:
Thanks for this contribution. By the way, your plugin's readme seems to suggest that there is a `list` command, however said command doesn't seem to work, and I see no code for it in the source

plugins/Env-0F43C966617C4817B1BEC29DCC8E5A69.json
```diff
@@ -0,0 +1,14 @@
+{
+    "ID": "0F43C966617C4817B1BEC29DCC8E5A69",
+    "ActionKeyword": "env",
```
cibere, line 3: This key should not be present here

plugins/Env-0F43C966617C4817B1BEC29DCC8E5A69.json
```diff
@@ -0,0 +1,14 @@
...
+    "Name": "Env",
+    "Description": "Control Environment Variables from Flow Launcher",
+    "Author": "lurebat",
+    "Version": "1.0.0",
+    "Language": "csharp",
+    "Website": "https://github.com/lurebat/Flow.Launcher.Plugin.Env",
+    "ExecuteFileName": "Flow.Launcher.Plugin.Env.dll",
```
cibere, line 10: This key should not be present here

cibere (changes requested):
Just noticed some extra keys in your file. Once these are removed, this can be merged

cibere (approved):
Looks all good to me now, thanks for removing those keys.

PR #529, opened 2025-07-04
add youdao-translate plugin

cyberdoors:


plugins/Youdao-translate-922d88e9-eee4-4d7e-89c7-eb916bbf65fd.json
```diff
@@ -0,0 +1,12 @@
+{
+    "ID": "922d88e9-eee4-4d7e-89c7-eb916bbf65fd",
+    "Name": "有道翻译",
```
cibere, line 3: The plugin's name here must match the name in this file's name.

So, if you wanted to keep the name "有道翻译", this file should be renamed to `有道翻译-922d88e9-eee4-4d7e-89c7-eb916bbf65fd`. While if you want to keep the name in the filename (Youdao-translate), then you should change the name key here to `Youdao-translate`

cibere (changes requested):

jjw24 (changes requested):
Hi thank you for the plugin. Unfortunately the filename will need to be in english. The description can be any language.

PR #530, opened 2025-07-07
Add MDict plugin

Jack251970:
Search in MDict dictionaries

image


cibere (approved):
ooh, that use of the preview panel looks cool 👀

PR #535, opened 2025-07-12
 Create HowLongToBeat-1e6b5249-32da-46e9-95e1-1b139f3f0ffc.json

Jack251970:
From @git-psi:

I created a new HowlongTobeat plugin because the old one I found does not work


cibere:
Filename must have the correct name

PR #537, opened 2025-07-24
Create rip-2F2C2193E9A54437AA69B846121976D8.json

iivinn:


cibere (changes requested):
Everything looks good to me, however it looks like despite having a release workflow, the plugin's releases were manually made and the workflow wasn't used. Please use the workflow to create the releases, or at least to upload the zip file.

PR #542, opened 2025-08-06
Updated (Enhanced) Direct Translate Flow Launcher Plugin

SweedXD:
### 🌟 Key Features
*   **Configurable Default Language:** Set your favorite language as the default.
*   **100+ Language Support:** Access the full Google Translate library.
*   **Command-Driven Interface:** Use `tr list` and `tr set <code>` to manage settings.
*   **Smarter Translation Logic:** The plugin is better at understanding your queries.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -3821,19 +3821,5 @@
         "IcoPath": "https://cdn.jsdelivr.net/gh/pivotiiii/flow_launcher_timer@main/img/app.png",
         "LatestReleaseDate": "2025-06-23T22:17:42Z",
         "DateAdded": "2024-07-07T09:19:19Z"
-    },
-    {
-    "ID": "a74621e00ca34dfea26b4aa7a612834e",
-    "ActionKeyword": "tr",
-    "Name": "Direct Translate (Enhanced)",
-    "Description": "Translate between any languages supported by googletrans",
-    "Author": "SweedXD",
-    "Version": "3.0.0",
-    "Language": "python",
-    "Website": "https://github.com/SweedXD/Direct-Translate",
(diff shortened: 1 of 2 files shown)


SweedXD:
ok, no problem

cibere (changes requested):
1. Please use the [github action](https://github.com/SweedXD/CurrencyPP/blob/master/.github/workflows/publish.yml) to create and package the release version of the plugin.
2. The plugin's name is still set as `CurrencyPP (UI Fixed)` in its [`plugin.json`](https://github.com/SweedXD/CurrencyPP/blob/master/plugin.json) file

cibere:
> Sorry for bothering you, I'm new to this. I updated the [`plugin.json`](https://github.com/SweedXD/CurrencyPP/blob/master/plugin.json) file. Now what should I do, I didn't update the Release yet

For releases, we require the plugin itself to be built and published by a github action. The previous developer of CurrencyPP setup [this workflow](https://github.com/SweedXD/CurrencyPP/blob/master/.github/workflows/publish.yml) to do just that, so you should be able to just re-use the same one. Though, to me at least, it appears that github doesn't recognize the workflow. I don't remember the "correct" way to fix this, however I believe pushing a commit to delete it, then another commit to re-create it should work. As far as how to use the workflow, it appears to have two triggers:
```
on:
  workflow_dispatch:
  push:
    branches: [ master ]
```
`workflow_dispatch` lets you manually trigger the workflow via the actions tab in the repository, and `push` makes the workflow automatically run upon receiving a commit. So, simply ensuring that github recognizes the workflow *should* be enough to get it to properly trigger.

cibere (approved):
Everything looks fine now, thanks for making all of those changes.

PR #546, opened 2025-08-25
Add docker plugin

thisisleobro:
This plugin allow users to list images and containers from Flow Launcher. We should add more features soon


cibere (approved):
Everything here looks good to me, however since i don't personally use docker, I'm going to have another team member manually test it with flow.

cibere:
@deefrawley confirmed that it works, so I'll merge this now

PR #554, opened 2025-09-14
Add new plugin: Insert 

llxee:
A plugin that allows users to replace placeholders in the template string


cibere (changes requested):
As it stands now, the latest version of your plugin from the releases tag is unusable. Just like how to changed the `Language` value here, you also need to change the `language` value in your plugin's `plugin.json` file, then create a new release. 

Other than that, everything else about this plugin looks just fine to me.

PR #566, opened 2025-11-08
Add FlowGet plugin.

benello:
Been using for FlowLauncher for a while and noticed that there is no plugins that integrate winget directly into it. So I gave it a try. This was inspired by the scoop plugin which is quite similar.


cibere:
> Looks like there’s already another MR for the same thing xD. I’ll keep it here for now. I’d love to add the plugin and help maintain it wherever possible.

I believe this is actually the third one written in csharp, however the one currently on the manifest only supports updating. Unless you have a specific reason for having a third plugin for winget, I will probably be closing this PR.

benello:
Well this aimed on adding more functionality such as searching and installing new packages while also adding the option to update or uninstall existing packages.

PR #567, opened 2025-11-08
Jira Search Version 1.2.0

haugjan:
- Added new query operators: ? for "In Progress" status and @? for unassigned issues
- Implemented 300ms debounce delay to reduce unnecessary API calls during typing
- Changed logging from LogInfo to LogDebug for reduced verbosity
- Refactored test helper class into a separate file for better code organization


cibere:
Plugins automatically update, so there is no need to manually PR updates

PR #573, opened 2025-12-02
Add Zed Workspaces plugin configuration file

shubhambhatnag:
Adding the configuration file for my Zed Workspaces extension: https://github.com/shubhambhatnag/zed-flow-extension


shubhambhatnag:
Updated my plugin.json - https://github.com/shubhambhatnag/zed-flow-extension/blob/main/plugin.json

PR #574, opened 2025-12-03
Plugin Submission: FlowTyper

nebbyJammin:
Implemented a simple, but customizable typing test within the flow launcher search bar.


cibere (changes requested):
1. The `Website` key in your `plugin.json` file leads to a non-existent repo
2. Your `plugin.json` file has two keys that it should not have: `UrlDownload` and `UrlSourceCode`
3. The value for `PackageProjectUrl` in `nebbyJammin/FlowTyper/Flow.Launcher.Plugin.FlowTyper/Flow.Launcher.Plugin.FlowTyper.csproj` has a typo
4. Your readme file mentions a settings/config file system that is separate from the one flow provides. If you are indeed doing that, please switch to using the one that flow provides (unless you have a legitimate reason for doing otherwise). If you aren't, then please update your readme accordingly.

PR #583, opened 2025-12-18
Add FlowFaker plugin

t4nk3rchu:
Hi, 
I'd like to add my FlowFaker plugin to the Plugin Store. This plugin would help user generate fake data using up to date FakerJS library directly

There's already a plugin Fake Data that use bogus library but it doesn't support good enough for eastern countries


cibere (changes requested):
Please remove the `node_modules` dir from your repo, and build them in your publish action instead

PR #607, opened 2026-02-09
Add Virtual Desktops plugin

DavidGBrett:
Add plugin that allows switching of virtual desktops using flow launcher.


deefrawley:
Still getting invalid:

image

Try something like https://www.uuidgenerator.net/version4

deefrawley:
Not sure what has gone before but if you could change it to remove all doubt that would be good

PR #612, opened 2026-02-18
Add Command Vault plugin

BossFlap:
## Command Vault

A 1Password-style command launcher for network and system admins.

- **Author:** BossFlap
- **Language:** Python
- **Keyword:** `cv`
- **Source:** https://github.com/BossFlap/flow-command-vault
- **Download:** https://github.com/BossFlap/flow-command-vault/releases/download/v1.0.0/CommandVault-1.0.0.zip

### Features
- 156 built-in commands: Cisco IOS, Linux, Proxmox, Ansible
- Full-text search (FTS5) with LIKE fallback
- Favorites system
- Template variables with interactive prompts
- Built-in GUI manager (`cv :manage`)
- SQLite backend, MIT license


deefrawley:
We generally discourage plugins where users need to do external actions outside of Flow to get it working (you will also find a lot of support tickets because people don't always follow READMEs). Can you add the db_init function as a keyword command with help text to guide the user if they don't do that and try and use your plugin?

PR #629, opened 2026-03-02
Add Multi-Profile Bookmarks plugin

sergebat:
This plugin opens bookmarks in correct browser profile. Ideal if you use separate profiles for work/study and personal browsing. Only Chrome is supported in this initial version. 

image


jjw24:
Hi @sergebat, all looks ok to me, but I think it might be worth updating the description to indicate that it currently supports Chrome only?

jjw24 (approved):

sergebat:
>  but I think it might be worth updating the description to indicate that it currently supports Chrome only?

Oh, sorry, missed this request @jjw24 . I agree and will fix the description asap. 

I hope I can make a first pass on Firefox support soon.

PR #645, opened 2026-04-11
Add Browser History V2 plugin

fahim-ahmed05:
This started as a fork of the [Browser History](https://github.com/Garulf/Browser-History) plugin, but I ended up rebuilding pretty much everything from scratch. The goal was to make it faster and just easier to use overall. 

I opened a [PR to the Browser History](https://github.com/Garulf/Browser-History/pull/37) repo two years ago, but it was never merged or rejected, and the project hasn’t had any updates since. Because of that, I decided to submit my version to the store.

### What’s improved

* **Better multi-word search**
  You can type words in any order and still get the right result. For example, `github flow` will match "Flow Launcher Plugin - GitHub" by checking both the title and URL.

* **Copy to clipboard**
  Open the context menu with `Shift + Enter` to quickly copy a URL without opening it.

* **Broader browser support**

  * Chromium-based: Chrome, Edge, Brave, Brave Nightly, Vivaldi, Arc, Thorium, Helium
  * Firefox-based: Firefox, Firefox Nightly, Zen, Floorp, Waterfox
  * Opera supported separately

* **All Browsers mode**
  Search history from all supported browsers at once.

* **Auto-detect active profile**
  Automatically uses the profile you most recently interacted with, like Work or Personal.

* **Search across all profiles**
  Finds and searches every profile inside a browser dynamically.

* **Custom profile support**
  Point it to any browser profle folder, including portable setups. It detects whether it's Chromium or Firefox and handles it automatically.

* **Domain blocking**
  Filter out sites like `facebook.com` or `youtube.com` so they don’t show up in results.

* **Live database caching**
  Safely copies active SQLite databases to avoid file lock issues while browsers are running.

* **Profile isolation via hashing**
  Prevents cache conflicts when switching between custom profiles.

* **Memory optimizations**
  Uses a configurable history limit and pushes filtering into SQLite queries to avoid lag, even with large histories.

### DIFF
--- a/plugins/Browser History V2-2032fd77-87f3-47fa-91ab-a67982ba1253.json
+++ b/plugins/Browser History V2-2032fd77-87f3-47fa-91ab-a67982ba1253.json
@@ -1,0 +1,12 @@
+{
+    "ID": "2032fd77-87f3-47fa-91ab-a67982ba1253",
+    "Name": "Browser History V2",
+    "Description": "Search your internet browsing history across multiple browsers",
+    "Author": "Fahim Ahmed",
+    "Version": "1.6.6",
+    "Language": "python",
+    "Website": "https://github.com/fahim-ahmed05/browser-history-v2",
+    "UrlDownload": "https://github.com/fahim-ahmed05/browser-history-v2/releases/download/v1.6.6/browser-history-v2.zip",
+    "UrlSourceCode": "https://github.com/fahim-ahmed05/browser-history-v2/tree/main",
+    "IcoPath": "https://cdn.jsdelivr.net/gh/fahim-ahmed05/browser-history-v2@main/icon.png"
+}
--- a/plugins/Browser History V2-2032fd77-87f3-47fa-91ab-a67982ba1253.json
+++ b/plugins/Browser History V2-2032fd77-87f3-47fa-91ab-a67982ba1253.json
@@ -3,10 +3,10 @@
     "Name": "Browser History V2",
     "Description": "Search your internet browsing history across multiple browsers",
     "Author": "Fahim Ahmed",
-    "Version": "1.6.6",
+    "Version": "1.6.7",
     "Language": "python",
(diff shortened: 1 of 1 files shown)


jjw24:
If it's rewritten from scratch, do you want to call it a different name or happy with one?

fahim-ahmed05:
> If it's rewritten from scratch, do you want to call it a different name or happy with one?

I’ve honestly lost track of how much of it I’ve rewritten at this point. I started just adding features, but that turned into reworking whole sections, so it’s kind of a mix of old and new now. As for the name… yeah, naming isn’t really my strong suit either 😄 I’m fine keeping the current one.

fahim-ahmed05:
@jjw24 do I need to change anything for approval?

jjw24:
I haven't had a chance to review & test the plugin yet, will do so soon.

jjw24:
@fahim-ahmed05 I would like to propose that the code change the location of where the cache copy of the browser database lives. Temp directory usually is for this sort of functionality but leaving the user's browser data there permanently is a slight privacy concern. I think a better place might be in the plugin's settings folder inside UserData location, because at least if flow is completely uninstalled/removed the cache is also gone. You can also use the computer name to invalidate cache via the `init` function for the user scenario where flow is moved to a different computer via portable mode.

A secondary thing is please indicate in the readme for users that a cached copy of the browser data is used and their stored locations.

Let me know what you think, happy to discuss further.

fahim-ahmed05:
@jjw24 Thank you for your feedback. I implemented your proposed changes. https://github.com/fahim-ahmed05/browser-history-v2/compare/v1.6.6...v1.6.7

jjw24 (approved):
@fahim-ahmed05 one final thing, in the README's privacy section please make a warning/note that if you use cloud sync with flow, the browser history database cache will also be uploaded.

fahim-ahmed05:
@jjw24 I've updated the readme with a warning. thank you for approving.

PR #646, opened 2026-04-11
Add zoxide plugin

WantChane:
Zoxide integration for Flow Launcher, written in C#.

PR #650, opened 2026-04-27
Remove ZoxidePy Plugin 

WantChane:
zoxidepy was archived


deefrawley:
Why is this being removed so soon after being added as a plugin?

jjw24:
Hello, it's re-written as a separate plugin so this should be de-listed.
https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/pull/646#issuecomment-4313605773

jjw24 (approved):

PR #652, opened 2026-05-01
Fix UrlSourceCode for Jira Search and Confluence Search to unblock auto-updater

haugjan:
## Summary

Removes the `.git` suffix from `UrlSourceCode` for two plugins by the same author:

- Jira Search (ID `BD32A62C-6F98-4541-AE8E-17B46458595F`)
- Confluence Search (ID `EE691863-50D1-4E0F-9A36-83314C2B2B27`)

## Why

`ci/src/updater.py` derives the GitHub API repo path from `UrlSourceCode`:

```python
url_parts = info[url_sourcecode].split("/")
repo = "/".join(url_parts[3:5])
```

With a value like `"https://github.com/haugjan/Flow.JiraSearch.git"`, this yields `repo = "haugjan/Flow.JiraSearch.git"`, which makes the request resolve to `https://api.github.com/repos/haugjan/Flow.JiraSearch.git/releases/latest` — that endpoint returns 404. The 404 path is not handled explicitly: `latest_rel.get("assets")` returns `None`, the `if assets:` block is skipped, and the function returns the plugin entry unchanged. No exception is raised and no log line is written, so this fails silently.

Symptoms in the manifest for both entries:

- `LatestReleaseDate` is `null` (the updater would otherwise set it on the first successful run)
- `etags.json` holds an empty ETag for the plugin's ID

For Jira Search the manifest is also stuck at `Version: 1.1.0` while the source repo is at `v1.2.0` (published 2025-11-08). Confluence Search happens to be in sync only because its entry was added (2025-11-18) after the latest release was published — any future release would silently be skipped too.

After this change, the next scheduled `Auto Update` run should pick up v1.2.0 for Jira Search and set `LatestReleaseDate` for both entries.

## Possibly worth a follow-up (separate)

It might be worth making `updater.py` strip a trailing `.git` from `UrlSourceCode` before splitting, and/or surfacing non-200/304/403 responses as warnings, so the same problem in other entries does not silently sit there. Happy to file a separate issue if that helps.

## Test plan

- [x] Diff is two single-line `UrlSourceCode` fixes, one per plugin.
- [ ] After merge, the next `Auto Update` workflow run picks up `v1.2.0` for Jira Search and sets `LatestReleaseDate` for both entries.

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -4401,7 +4401,7 @@
         "Language": "csharp",
         "Website": "https://github.com/haugjan/Flow.JiraSearch",
         "UrlDownload": "https://github.com/haugjan/Flow.JiraSearch/releases/download/v1.1.0/JiraSearch-1.1.0.zip",
-        "UrlSourceCode": "https://github.com/haugjan/Flow.JiraSearch.git",
+        "UrlSourceCode": "https://github.com/haugjan/Flow.JiraSearch",
         "IcoPath": "https://cdn.jsdelivr.net/gh/haugjan/Flow.JiraSearch@main/Flow.JiraSearch/Images/icon.png",
         "LatestReleaseDate": null,
         "DateAdded": "2025-10-22T01:53:34Z"
--- a/plugins.json
+++ b/plugins.json
@@ -2189,7 +2189,7 @@
         "Language": "csharp",
         "Website": "https://github.com/haugjan/Flow.ConfluenceSearch",
         "UrlDownload": "https://github.com/haugjan/Flow.ConfluenceSearch/releases/download/v1.0.1/ConfluenceSearch-1.0.1.zip",
-        "UrlSourceCode": "https://github.com/haugjan/Flow.ConfluenceSearch.git",
+        "UrlSourceCode": "https://github.com/haugjan/Flow.ConfluenceSearch",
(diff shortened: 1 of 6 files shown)


haugjan:
Hi @cibere, sorry for the direct ping! Whenever you have a quiet moment, would you mind taking a look at this PR when it suits you? It's a one-line fix per entry — removing a `.git` suffix from `UrlSourceCode` that's making the auto-updater hit a 404 and silently skip my two plugins. CI is green and the description has the details. No urgency at all, I just wanted to flag it once rather than let it sit unnoticed. Thank you so much for maintaining this — really appreciate everything you do here.

Jack251970 (dismissed):
Hi, thanks for your fix!

Please do not directly change `plugin.json` which is auto-updated in CI. You should change your plugin json file `plugins/Confluence Search-EE691863-50D1-4E0F-9A36-83314C2B2B27.json` and `plugins/Jira Search-BD32A62C-6F98-4541-AE8E-17B46458595F.json`. And CI will handle update in `plugin.json` automatically.

jjw24:
Hey, I will get this fixed, was in the middle of making changes anyway.

jjw24 (approved):
Thank you for investigation and thorough explanation of the issue 👍

jjw24:
We want to add a validation for UrlSourceCode to be in the same format so that the automation code is simpler. I will create an issue for this.

PR #673, opened 2026-06-04
fix: remove MinimumAppVersion from FMHY Search plugin

iamshamit:
Removes the `MinimumAppVersion: 2.12.0` restriction that was incorrectly excluding the plugin for users on Flow Launcher 2.1.x. The plugin works on all current versions.


Jack251970 (dismissed):

Jack251970 (approved):
Thanks for your contribution!

PR #679, opened 2026-06-11
Update ShareX_Flow_Plugin-75763350-A89D-4942-B697-14875D94C222.json

dufguix:
v1.1.0


deefrawley:
If your plugin has already been approved then you just need to publish an update and the store will pick it up. I can see v1.1.0 has been picked up so close this PR. Thanks

PR #681, opened 2026-06-15
Add Text Tools plugin

ghaemaghaey:
A Python plugin for Persian and general text manipulation.
Converts Arabic characters to Persian, removes half-spaces, cleans lines and more.

https://github.com/ghaemaghaey/Flow.Launcher.Plugin.TextTools


ghaemaghaey:
mistake in the how should the approach toward the implementin was made

PR #684, opened 2026-06-16
Add VSCode Recent plugin

xt-fg:
Quickly find and open your recent VS Code folders and workspaces from Flow Launcher.


deefrawley:
Hi, there are currently several existing plugins to open VSCode workspaces. Have you reached out to those authors to see if you can improve one of those plugins, or what does your plugin do that is unique or enhanced?

xt-fg:
> Hi, thanks for the question. I have not contacted the existing plugin authors because this is an
  > independently developed implementation with a different scope and storage strategy, rather than a fork or
  > patch of an existing plugin.
  >
  > I reviewed the existing options:
  >
  > - VS Code Workspaces currently relies on the legacy SQLite storage location, and it has an open issue
  >   where no results are returned since VS Code 1.118:
  >   https://github.com/taooceros/Flow.Plugin.VSCodeWorkspace/issues/24
  >
  > - RecentProjectsOpen supports SSH and WSL, but it is a general multi-IDE launcher. Its VS Code
  >   implementation only reads profileAssociations.workspaces, and local and remote projects are exposed
  >   through separate providers.
  >
  > VSCode Recent differs in the following ways:
  >
  > - Reads profileAssociations, backupWorkspaces, and windowsState.
  > - Uses workspaceStorage timestamps for actual recency ordering.
  > - Shows local folders, workspace files, SSH, WSL, Dev Containers, Codespaces, and other remote URIs in one
  >   unified list.
  >
  > - Supports VS Code, VS Code Insiders, and VSCodium.
  > - Avoids depending on the live SQLite database and filters deleted local paths.
  > - Uses only the Python standard library, with no third-party dependencies.
  >
  > I have also corrected the documentation regarding VS Code 1.118: it moved history.recentlyOpenedPathsList
  > to shared application storage rather than removing the key entirely. This plugin intentionally uses
  > storage.json as its primary source, with the legacy database only as a fallback.
  >
  > Although there is some overlap with the existing plugins, I believe the broader storage coverage, unified
  > remote support, recency ordering, and dependency-free implementation provide meaningful differentiation.

deefrawley:
Update your submission to match your repo -1.0.1

jjw24:
Tested 1.0.1 works as intended.

jjw24 (approved):

PR #688, opened 2026-06-17
Add AudioCowboy plugin

True347:
Adds **AudioCowboy**, a Python plugin to quickly switch the default audio input/output device and save/load device profiles, all under one action keyword (`ac`) with a rofi-style drill-down menu.

- **ID:** `54dcf7b098df441ab7c2ad6b11c047cc`
- **Repo:** https://github.com/True347/AudioCowboy
- **Release:** [v1.0.0](https://github.com/True347/AudioCowboy/releases/tag/v1.0.0) (zip has `plugin.json` at root + vendored `lib/`)

Checklist:
- [x] `plugins/AudioCowboy-<id>.json` added; copied fields match the plugin's `plugin.json`
- [x] `UrlDownload`, `UrlSourceCode`, `IcoPath` all resolve (release zip + jsdelivr icon verified live)
- [x] GitHub Actions build/release workflow set up (auto-publishes on version bump)
- [x] Conforms with the Plugin Store policy

Switching sets all three roles (Console + Multimedia + Communications) so apps like Teams/Discord/Zoom follow. Ships a pure-Python `pycaw` backend, so no external binary is required.


deefrawley:
Hi, there are already several audio device changing plugins for Flow. Does this do something new or better or have you considered contributing to one of the others?

True347:
Thanks for taking a look. I did check the manifest before submitting — the only listed plugin I found that changes the default device is Audio Device Selector. The other audio entries (App Audio Manager, FlowTrumpet, VolumeFlow, MediaControl) are volume or playback controls.

Compared with it, AudioCowboy adds:

- **Input devices** — switches microphones, not just playback.
- **All three roles** — Console + Multimedia + Communications, so Teams/Discord follow.
- **Named profiles** — save an output+input pair and apply both in one action.
- **Endpoint IDs instead of device names** — two identical headsets stay distinguishable.
- **Verification** — it re-reads the device list after a switch and reports applied / unverified / failed rather than assuming success.

The first two are open requests there ([#19](https://github.com/attilakapostyak/Flow.Launcher.Plugin.AudioDeviceSelector/issues/19), [#20](https://github.com/attilakapostyak/Flow.Launcher.Plugin.AudioDeviceSelector/issues/20)); its last release was April 2023. Contributing there instead would be a rewrite rather than a patch — it's C# on .NET 5, this is Python.

Happy either way. Which would you prefer?

deefrawley:
Why are you downloading an external exe without hosting its source? This is frowned upon.

True347:
You're right — thanks for flagging it. That was an optional `svcl.exe` backend a `setup.ps1` helper could fetch; it was never in the release zip, but rather than keep a path that pulls a closed-source binary at all, I've removed it entirely in v1.0.4 — the downloader, the backend code, and the setting are all gone. The plugin now uses only pycaw and its open-source dependencies: nothing downloaded, no closed-source binary. The manifest here is updated to v1.0.4.

jjw24:
Reviewed svcl.exe no longer downloaded.

jjw24 (approved):

PR #695, opened 2026-07-02
Fix VirusTotal scan CI

jjw24:
Switch to PR target to allow scanning on fork PR branches. Follows on from #694


Jack251970 (dismissed):

.github/workflows/security-vt-scan.yml
```diff
@@ -4,32 +4,50 @@ on:
...
-        if: github.event_name == 'pull_request'
         id: filter
         with:
+          ref: ${{ github.event.pull_request.head.sha }}
           filters: |
             plugins:
               - 'plugins/**'
+            ci_files:
+              - 'ci/**'
+              - '.github/**'
```
DavidGBrett: Given that PRs to add a plugin should only make changes in the plugins folder,
Perhaps we should broaden this to all files outside that folder, instead of just ci_files,
Just to be safe.

so something like
```suggestion
            non_plugin_files:
              - '**'
              - '!plugins/**'
```

with ``non_plugin_files`` replacing other uses of ``ci_files``

.github/workflows/security-vt-scan.yml
```diff
@@ -4,32 +4,50 @@ on:
...
```
DavidGBrett, line 41: If I'm not wrong - a PR that modifies both plugin files and non plugin files would cause the virus scan to skip - which might be missed in a review

Perhaps we could add a test like this to explicilty fail if mixed changes like that are made

```suggestion

validate-pr-scope:
    needs: changes
    if: |
      github.event_name == 'pull_request_target' &&
      needs.changes.outputs.plugins == 'true' &&
      needs.changes.outputs.non_plugin_files == 'true'
    runs-on: ubuntu-latest
    steps:
      - name: Fail PRs that mix plugins and other changes
        run: |
          echo "This PR changes files in the plugins directory as well as files outside it."
          echo "Plugin PRs must only modify files in the plugins directory."
          exit 1

```

DavidGBrett (dismissed):
LGTM!

Those comments I made have been addressed, thanks!

jjw24:
They are good review comments; I have also further locked it down to allow only .json files in plugins directory.
Please review the new changes I added as well.

DavidGBrett (approved):
No further comments - looks good to me!

PR #705, opened 2026-07-10
Add LeetCode Search plugin

ErenMFYeager:
Adds plugin manifest for LeetCode Search — fuzzy search, topic/difficulty filters, daily challenge, and local caching for LeetCode problems.


ErenMFYeager:
Hi @jjw24, opened a fresh PR with just the plugin manifest file as suggested — should be clean now. Let me know if anything else is needed!

jjw24 (changes requested):
Hi there, it's missing lib directory for your imports, so getting this when using the plugin:
{0DFE8631-A3AF-44A9-B1B6-44F07C70834B}

ErenMFYeager:
Hey, sorry about that, my bad. Turns out the lib/ folder with all the dependencies was gitignored so it never actually made it into the build, and I'd also forgotten to add rapidfuzz to requirements.txt. Fixed both, bumped it to v1.4.3, cleaned up the release zip, and updated the manifest here to point to the right version and download link. Tested it on my own Flow Launcher and it's working now. Would you mind pulling the latest and giving it another shot?

deefrawley:
Hi, can you also ensure you aren't breaching leetcode terms and conditions with the way you grab the data in the background? Thanks

ErenMFYeager:
Hey @deefrawley, thanks for calling this out, it's a legitimate concern., The indexing worker is scraping LC's unofficial graphql endpoint pretty aggressively in the background, that's a legit ToS issue.

Here's what I am thinking of doing reduce the load on LC's servers:

- Gonna ship `problems_cache.json` bundled with every release instead of indexing on first run. So a fresh install just has the data already, no hitting their servers at all on install.
- I'll cut a new release roughly every 15 days to refresh that cache. Doing it manually on my end, not something that runs on a user's machine. New questions mostly drop on weekends anyway and it's like 4 questions, I doubt anyone's missing them for two weeks when the rest of the library is already huge.
- Rate-limiting `lc refresh` to once a day or maybe once a week so people can't just spam it and turn it back into a mini-scraper.

Let me know if that's enough or if you want more. Since I have to encounter my exams in few days, I won't be able to do major changes right now, but would love to figure out how to remove or reduce this issue.

deefrawley:
Hi @ErenMFYeager, sorry for the late reply. The leet code Use of Service in their Terms state

`Activities such as “crawling,” “scraping,” or “spidering” any part of the Service, and attempting to decompile, reverse engineer, or discover the source code or underlying ideas of the Service, are strictly forbidden.`

IANAL but based on that, getting the data to display rather than reverse-engineering is ok? Not having the client installs do it will definitely help. Probably best to ensure you are attributing everything properly back to them too.

PR #706, opened 2026-07-14
Add QuickWall plugin submission

yazan-abualenain:
QuickWall — browse wallpaper folders and set desktop wallpapers from Flow Launcher. Python plugin, MIT license.


Garulf:
Hi,

I noticed your plugin is using flox. Just a heads up: flox was deprecated a couple of years ago in favor of [pyFlowLauncher](https://github.com/Garulf/pyFlowLauncher), a newer, typed library that supports the plugin v1 AND v2 API. 🎉

I'd highly recommend migrating, flox has a few known flaws that won't be fixed due to how Flow Launcher has evolved, and pyFlowLauncher addresses them.

I also noticed flox is vendored in your project's source. That works, but it makes updating awkward. Most plugins instead have their CI bundle the dependency into the release zip artifact.

Thanks for using flox, and I hope you enjoy pyFlowLauncher!

PR #710, opened 2026-07-15
Update Steam Search description and icon path

Garulf:
Just updates the description and icon path for Steam Search plugin.


DavidGBrett (approved):
looks good to me!
small change like this is certainly safe

PR #712, opened 2026-07-16
Fix ci plugin changes detection

### LINKED ISSUE #726: VirusTotal scan failing due to missing scan-files dir
See test PR #725

```
FILE_ERROR: path(s) not found: ./scan-files/
  VirusTotal scan | paths=['./scan-files/'] | no_cache=False | cache=vt_cache.json | whitelist=vt_whitelist.json | VT_REQUEST_INTERVAL_SEC=15.0
```

The pull_request_target part isn't being blocked anymore, but it did still fail later on.

Could be because this is an edit not a new plugin submission - though either way we need to fix that.

Currently #712 & #728 are open to solve this

jjw24:
Fix the changes detection to detect only when new files added not just changes made to the plugin directory files.

Tested:
1. When only plugin files modified -> VT changes detect pass -> VT scan skipped
2. When files changed but not in plugin directory -> VT changes detect pass -> VT scan skipped
3. When files changed and also plugin directory files changed -> VT changes detect fail
4. When only plugin directory files modified but no new files added -> VT changes detect pass -> VT scan is skipped
5. When only new Json file added in plugin directory -> VT changes detect pass -> VT scan runs

### DIFF
--- a/.github/workflows/security-vt-scan.yml
+++ b/.github/workflows/security-vt-scan.yml
@@ -48,10 +48,11 @@ jobs:
           ADDED=$(git diff --diff-filter=A --name-only ${{ github.event.pull_request.base.sha }} HEAD -- 'plugins/*.json' | head -c1)
           if [ -n "$ADDED" ]; then
             echo "plugin_additions=true" >> $GITHUB_OUTPUT
+            echo "New plugin manifest file/s added: true"
           else
             echo "plugin_additions=false" >> $GITHUB_OUTPUT
+            echo "New plugin manifest file/s added: false"
           fi
-          echo "New plugin file/s added: $ADDED"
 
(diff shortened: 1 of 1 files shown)


DavidGBrett (commented):
I'm assuming this is aiming to solve the failure in #710

I'm not sure if this is the right approach..

If we don't scan on changes at all then someone could change the ``UrlDownload`` field to introduce malicious code.
I think we should be more careful about what changes are allowed to skip the virus check.

To make this work with scanning dangerous changes, I believe we would need to make the `download_plugins.py` script also download the plugin again so `./scan-files/` exists for the virus scan workflow.

DavidGBrett:
Closing this PR as after #728 (and #744) we now already have a working version of this

PR #713, opened 2026-07-22
Add Lumina plugin

bijanmurmu:
Submitting a new plugin for Flow Launcher. Lumina is a dynamic Windows customization tool that allows users to seamlessly switch Wallpapers, Light/Dark themes, and Accent colors directly from Flow Launcher. It features a custom hardware-accelerated WPF transition engine for smooth crossfade animations during wallpaper switching.


deefrawley:
Hi, you bundle an executable without source code which is generally frowned upon. Can you please provide the source or more information on what this file is.

bijanmurmu:
> Hi, you bundle an executable without source code which is generally frowned upon. Can you please provide the source or more information on what this file is.

The executable is a transition engine I wrote to create smooth animation when the user switches their wallpaper, instead of a sudden change, and for the source code it is there inside the `core` as `animator.cs` and `LuminaAnimator.csproj`.

Let me know if you need me to compile it differently or if this is not the preferred way to handle bundled executables.

jjw24 (changes requested):
Hi @bijanmurmu, looks like the C# app is core/LuminaAnimator.csproj and the current release workflow only zips up the repo. Right now, it does not build the C# project and this is the missing link.

You can change the workflow to:
1. build the C# project
2. copy the build output into the plugin/package directory
3. zip the final directory
4. publish

This way, you don't need to keep the exe binary, and your plugin is fully transparent.

bijanmurmu:
@jjw24 Thanks for the feedback, I've made the requested changes to the repository.

jjw24 (dismissed):
Thank you for making the changes.

jjw24:
After reviewing the animator.cs (LuminaAnimator.exe) code- whitelisted the detection from VT's engine DeepInstinct. False positive alert likely caused by the use of low-level Win32 window-manipulation APIs plus a transparent full-screen desktop overlay, which can resemble shellcode or unpacker behavior to heuristics.

jjw24 (approved):

PR #723, opened 2026-08-03
Add Ai Assistant using Groq plugin

Omprakash-Wagh:


Omprakash-Wagh:
Hi team! Just a heads up that the VirusTotal check failed immediately with a pull_request_target checkout error (Refusing to check out fork pull request code...). It looks like the GitHub Action is blocking the checkout from my fork to protect repository secrets. Let me know if you need anything from me once the workflow is adjusted or manually run!

jjw24 (dismissed):
Hi @Omprakash-Wagh, thank you for your submission. There is already a plugin called AI Assistant, could you please change your plugin name- since it uses Groq API, perhaps "AI Assistant for Groq"? Also please publish your releases via GitHub Action.

deefrawley:
Has this been done? JSON still shows as "AI Assistant". Reply here please

Omprakash-Wagh:
image_2026-08-24_111501094
I have, look at the image, i have updated the file name and the contents as well

Omprakash-Wagh:
Hi! I've added the required .json file for my plugin. However, the CI pipeline is failing on the download_plugins.py step with a git diff 128 error (it can't find origin/main). This seems to be an issue with the GitHub Actions checkout step for PRs from forks.

Omprakash-Wagh (commented):
ok

DavidGBrett:
Hi,
the CI bugs should be fixed now!

Though I noticed you included another plugin submission in this PR - could you separate those so each can be reviewed separately and clearly?

We prefer each plugin to have their own PR

Omprakash-Wagh:
@DavidGBrett  i have removed the other plugin file , please check

DavidGBrett (changes requested):
I tried installing this from the release zip but got this error 
`ModuleNotFoundError: No module named 'pydantic_core._pydantic_core'``
indirectly from `from groq import Groq`

It seems there is a problem with how your libs are being generated

Looking at the codebase it seems you hardcode the path to your installed plugins folder
The plugin folders name won't be `Ai Assistant` - it will be something like `Ai-Assistant-using-groq-1.1.0` etc which changes with version number
You also need to make sure that this works with the portable version of flow launcher

I'm not sure why you are trying to duplicate the api-key there but it would be better to look into storing it using something like the keyring package, and ideally clearing it from the user's settings after that or prompting them for it in the query window instead.

Your readme is also out of date with the new name for the plugin, and has incorrect instructions like typing `reload_plugins`

Please make sure that your plugin works using the release zip that you publish, and give everything a manual review

Omprakash-Wagh:
@DavidGBrett  i have made changes, it was because the github action that was causing the issue with the zip, i have fixed that and also updated the readme and implemented keyring package for proper pathing, please check

DavidGBrett (changes requested):
Same dependency issue again...
I looked into it and it seems pydantic_core is just a problematic dependency - it only works with the exact version of python specified - you have it for python 3.11 while flow on my system is using python 3.13
This is an odd scenario so I'll confirm with the team the best approach here and see if this is a blocker

Some other issues beyond that
The name of the plugin in your plugin.json would look better without the -'s and in "Title Case"

You hardcoded the action keyword `ai` - you need to get that each time as the user can change that

Other than that it does work correctly!

DavidGBrett:
Closed and re-opend to re-trigger the virus scan, which seemed to have an api outage - tests look green now

Omprakash-Wagh:
@DavidGBrett  i have made changes and also removed the groq dependency which was causing the issue with the pydantic_core binaries, please check now

DavidGBrett (changes requested):
Thanks for the fixes
Should be almost there now

The name in your plugin.json reverted to just AI Assistant without the using groq part

You also reference https://github.com/Omprakash-Wagh/flow-groq in your readme which doesn't exist

Omprakash-Wagh:
@DavidGBrett done, please check

DavidGBrett (approved):
That Option 2: Manual Git Installation section in the readme is still referring to `https://github.com/Omprakash-Wagh/flow-groq.git`
But I won't block this over a readme you can fix that later

Thank you for your contribution!

Omprakash-Wagh:
@jjw24  can you review the pull request and merge it?

Issue #719, opened 2026-07-30
Virus check failing due to refusal to check out fork pull request code from a 'pull_request_target' workflow


DavidGBrett:
All the recent plugin submissions are failing as github actions is refusing to use a workflow with a pull request target
E.g. https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/actions/runs/30347640779/job/90237497154?pr=716
This error message is given

"""
Error: Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access. Fetching and executing a fork's code in that trusted context commonly leads to "pwn request" vulnerabilities. To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.
"""

The change that caused this was in #695

jjw24:
The pull_request_target block is likely a recent change that has taken effect. Our current design includes safety mechanisms built in to run the pull_request_target context only when the expected plugin manifest files and nothing else, are added- minimizing the risk. GitHub's scanning does not recognise this, which is why it is still flagged as unsafe.

I do not see any increased risk mitigation benefit in reworking this workflow, but I do think we need to consider that while the safety mechanism currently works, it is prone to accidental changes that might render it ineffective in future edits, and mistakes could occur as this workflow evolves. I will make the changes to avoid this from happening,

jjw24:
I have tried several refactors, but they make the CI code messier and more convoluted than necessary, with no increase in security measures and the same risks still apply. I have opted in the PR to add the override option and some additional refinements to the security measures currently in place.

PR #724, opened 2026-08-03
Fix 'pull_request_target' workflow & allow modifiable vt_whitelist.json in plugin submission & additional security measures 

jjw24:
- Add `allow-unsafe-pr-checkout: true`
- Allow vt_whitelist.json to be modified in plugin submissions
- Pin GH actions to commit hash instead of tag version
- Add plugin download URL validation to ensure it's from GH and is zip extension

Closing #719


.github/workflows/security-vt-scan.yml
```diff
@@ -32,6 +33,8 @@ jobs:
             non_plugin_files:
               - '**'
               - '!plugins/*.json'
+              - '!vt_whitelist.json'
+      # NOTE: this is a security check to prevent non-plugin files from being changed. This check must remain in place and effective.
```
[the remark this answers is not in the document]
jjw24: This is intended- some plugins could get falsely flagged by a few VT engines. Add to whitelist is also transparent in the PR so the reviewer will be aware of the whitelisting.

.github/workflows/security-vt-scan.yml
```diff
@@ -14,31 +14,69 @@ jobs:
...
+              return
+            }
+
+            await github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              body: `${marker} This PR modifies vt_whitelist.json. Please review the whitelist addition before merging.`
+            })
+            console.log('Posted whitelist warning comment')
```
DavidGBrett: I'm a bit worried that a whitelist change might not be obvious enough, given the risk
so I added a step that should post a comment in the PR when the whitelist is in the changes

DavidGBrett (dismissed):
Disabled auto-merge so that you can check that warning step I added before merging.

Outside of the whitelisting it all looks good to me.
Good changes to check the download paths

Might be worth leaving the version pins only for the first party github actions (e.g. for dependabot to work) - though this is a debatable issue so I'll leave it up to you

.github/workflows/security-vt-scan.yml
jjw24: I think this might over-engineering it- as a standard benchmark for a malicious result you would see 10+ threat engines flagging it (yes there are also could be the case where new threats could only be detected by 1 or 2 engines), and one would need to list all 10+ threats in the whitelist for the plugin. So, a reviewer would catch that the whitelist is maliciously used.

DavidGBrett: I was more concerned that a review might not be careful enough with the whitelist changes specifically, thinking a reminder was no harm.
Though yes perhaps this is overkill (and extra complexity) - I can remove this then.

DavidGBrett (approved):
Removed my script to post a warning when whitelist is modified, so as to avoid excessive complexity.
I can add that back in a separate PR if we think it is worth it later again.

No other changes since then so I'll approve again.

PR #728, opened 2026-08-07
Add changed mode to download_plugins script

### LINKED ISSUE #726: VirusTotal scan failing due to missing scan-files dir

DavidGBrett:
Added new mode to download_plugins script - "changed"
It uses git diff to include both modified and added plugins

Made security-vt-scan.yml‎ use this mode instead of "new"

resolves #726

### DIFF
--- a/.github/workflows/security-vt-scan.yml
+++ b/.github/workflows/security-vt-scan.yml
@@ -80,9 +80,9 @@ jobs:
           key: plugin-zips-restore-not-used
           restore-keys: plugin-zips-
 
-      - name: Download plugin submitted in this PR
+      - name: Download plugins changed in this PR
         if: github.event_name == 'pull_request_target'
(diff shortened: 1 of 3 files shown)


DavidGBrett:
I think we could also remove the "new" mode as I don't think its used anywhere but I left it in for now

We should really be using the "changed" mode in all places from now on.

deefrawley (approved):

PR #731, opened 2026-08-09
Add GeminiSearch plugin configuration file

MiguellBorgess:


deefrawley:
Hi @MiguellBorgess - this plugin isn't working for me. It fires up the Gemini web site and copying the query to the clipboard but isn't then pasting into the web page. I'm running Win10

MiguellBorgess:
> Hi @MiguellBorgess - this plugin isn't working for me. It fires up the Gemini web site and copying the query to the clipboard but isn't then pasting into the web page. I'm running Win10

Hi! Thanks for the observation. But that's actually what's supposed to happen.

deefrawley:
Sorry then what is the purpose of this plugin?

PR #742, opened 2026-08-21
Add ObjectFlowed Plugin

FunnyDudeislazy:
Adds ObjectFlowed, a Flow Launcher plugin for opening Objectified episodes.


deefrawley:
Plugin isn't working for me. Type the keyword and nothing happens. You may want to put some exception checking and catching around the urlopen call to tell the user what is happening (and to stop from crashing Flow if there is something wrong with the URL)

PR #745, opened 2026-08-25
Update Playnite icon path

Garulf:
The Playnite icon moved to `data/icon.png` when I rewrote the plugin for v3.0.0, so the current `IcoPath` points at a file that is no longer in the repo root. raw.githubusercontent already 404s on it and jsdelivr is still serving a cached copy, so the store icon will break once that cache expires.

This just points it at the new path. Everything else in the entry is the same.


DavidGBrett (approved):
That virus check bug was fixed so everything should be good now

